PR #3, opened 2015-10-03
SmartDashboard command putter

aschokking:
This is ported from the 2015 repo's pull request.

Basically when we wanted to put commands on the smart dashboard (like certain reset/calibration commands that don't need buttons) we have now written code that can't be unit tested since the SmartDashboard static would break.

This add a helper function on commands you can call to set it to the smartdashboard if the code is running on the real robot or not.

@JohnGilb @WasabiFan @shahronak @riverole


review-ninja:
<a href="https://app.review.ninja/Team488/SeriouslyCommonLib/pull/3" target="_blank">ReviewNinja

JohnGilb:
This looks good!

PR #4, opened 2015-10-03
Clear the permanent store before saving properties

JohnGilb:
@aschokking @WasabiFan @riverole @shahronak 

This is something I saw happen a few times last year - where we would remove a property from the code, but then it would continue to get loaded/saved indefinitely until I deleted it manually from the flat file.

### DIFF
--- a/src/xbot/common/properties/SmartDashboardTableWrapper.java
+++ b/src/xbot/common/properties/SmartDashboardTableWrapper.java
@@ -17,6 +17,10 @@ public class SmartDashboardTableWrapper implements ITableProxy {
     public void setDouble(String key, double value) {
         SmartDashboard.putNumber(key, value);
     }
+    
+    public void clear(){
+    	// Do not clear the smart dashboard
+    }
 
     public Double getDouble(String key) {
         try {
--- a/src/xbot/common/properties/SmartDashboardTableWrapper.java
+++ b/src/xbot/common/properties/SmartDashboardTableWrapper.java
@@ -4,6 +4,8 @@
  */
 package xbot.common.properties;
 
+import org.apache.log4j.Logger;
+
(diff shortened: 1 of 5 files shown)


review-ninja:
<a href="https://app.review.ninja/Team488/SeriouslyCommonLib/pull/4" target="_blank">ReviewNinja

src/xbot/common/properties/SmartDashboardTableWrapper.java
```diff
@@ -17,6 +17,10 @@
     public void setDouble(String key, double value) {
         SmartDashboard.putNumber(key, value);
     }
+    
+    public void clear(){
+    	// Do not clear the smart dashboard
```
aschokking: why not?

aschokking: or throw?

tests/xbot/common/properties/PermanentStorageProxyTest.java
```diff
@@ -42,6 +42,20 @@ public void testSaveAndLoad() {
         assertEquals("What time is it?", p.getString("phrase"));        
     }
     
+    @Test
+    public void testClear() {
```
aschokking: This test makes me a little weary since it could pass even if clear didn't work but the write to file/load from disk part wasn't working.

JohnGilb: Added some extra validation after loading to make sure there's some data.

src/xbot/common/properties/SmartDashboardTableWrapper.java
JohnGilb: Added a warning line.

PR #5, opened 2015-10-03
Add persistence type to a property

riverole:
@aschokking @JohnGilb @WasabiFan 
Adding persistence type when creating a new property. This allows us to determine if a property will be saved to the disk when we persist the variables. I have created an enum PropertyPersistenceType with Ephemeral and Persistent labels, and created new constructors with this new persistenceType parameters. The old constructors will be deprecated in the future, I have added the deprecated tag to the old constructors. The default behavior when a "deprecated" constructor without persistence type is not used is persistent (conservative).


review-ninja:
<a href="https://app.review.ninja/Team488/SeriouslyCommonLib/pull/5" target="_blank">ReviewNinja

src/xbot/common/properties/DoubleProperty.java
```diff
@@ -28,10 +34,12 @@ public void set(double value) {
     }
     
     /**
-     *
+     * We only save the property if it's from a persistent type
      */
     public void save() {
+    	if(persistenceType == PropertyPersistenceType.Persistent) {
         permanentStore.setDouble(key, randomAccessStore.getDouble(key).doubleValue());
```
aschokking: tab in?

src/xbot/common/properties/PropertyManager.java
```diff
@@ -102,17 +103,48 @@ public void saveOutAllProperties() {
         permanentStore.saveToDisk();
     }
     
+    /**
+     * @deprecated You should use createProperty(String key, Boolean defaultValue, PropertyPersistenceType persistenceType)
```
aschokking: Instead of deprecating these factory functions maybe it would be friendlier to have like createProperty and createEphemeralProperty or something so it's a little leaner?

tests/xbot/common/properties/PropertyTest.java
```diff
@@ -65,6 +66,57 @@ public void testSavingValue() {
...
+        
+        assertSame(null, propertyManager.permanentStore.getDouble("weight"));
+        assertSame(null, propertyManager.permanentStore.getBoolean("isFalse"));
+        assertSame(null, propertyManager.permanentStore.getString("robotname"));
+        
+        assertSame(null, propertyManager.permanentStore.getDouble("height"));
+        assertSame(null, propertyManager.permanentStore.getBoolean("isAwesome"));
+        assertSame(null, propertyManager.permanentStore.getString("team"));
+        
+        assertEquals(0.5, dbl1.get(), 0.001);
```
WasabiFan: Why is the margin of error here `0.001`? Does the database system truncate to three places, or is this just because you're required to give a range?

src/xbot/common/properties/DoubleProperty.java
WasabiFan: There is a real tab character there; should make sure to eradicate those. @riverole Did you set up the checkstyle and formatting policies in your eclipse installation yet?

src/xbot/common/properties/Property.java
```diff
@@ -22,17 +22,42 @@
 
     ITableProxy permanentStore;
     ITableProxy randomAccessStore;
-
+    
+    /** 
+     * New enum to determine property persistence
+     */
+    public enum PropertyPersistenceType{
```
WasabiFan: Should probably add Javadoc comments to each enum value to explain what they mean.

tests/xbot/common/properties/PropertyTest.java
riverole: To be honest I didn't even gave a thought about this, I just copied over the tests from the normal save, and it was using that margin of error. Maybe @JohnGilb or @aschokking can comment on why we use a 0.001 margin of error.

aschokking: It's deprecated not to provide a margin with the double assertEquals and I think early on we chose this for most of our robot unit tests reasonable arbitrarily. Could probably be a constant or something on the base class most of the time.

src/xbot/common/properties/Property.java
aschokking: Not a bad idea, also maybe to the factory methods that create them?

JohnGilb: Yeah,I agree with Alex - factory methods would be pretty readable and autocomplete nicely.

PR #7, opened 2015-10-18
Namespace updates

aschokking:
A huge reshuffling of namespaces to be a little bit more sane. Basic idea:
* All mock classes should be in the test folder, they would never go into the robot jar
* All sensors/mechanisms/etc.. should be in the common.controls namespace

@WasabiFan @JohnGilb


aschokking:
github doesn't really show the renames well in this diff so might be worth checking out the branch locally to see how things look in eclipse.

WasabiFan:
I recall that that there were instances where we used mocks in either other places in the common lib or in the 2015 robot as placeholder objects if something didn't load correctly. You should try deploying to the robot to see if you get a ClassNotFound exception.

WasabiFan:
Are there any better words to replace "controls"? Maybe separate "sensors", "motors", etc.? I don't think that the name "controls" is particularly clear (I couldn't come up with much else).

aschokking:
That's a fair point about fall-back mocks, I'll do a scan of the 2015 project.

I'm open to better names for things, I agree controls isn't ideal but there's something that potentially envelops all of these.

aschokking:
Good call @WasabiFan, we did need MockGyro and MockRobotIO in src. Should be all good now though.

PR #11, opened 2015-11-07
Feature/checkstyle fixes

aschokking:
Fixed just about all of the checkstyle issues in the codebase (including adding in a no tabs rule).
Added a supression for the nav6 code which we don't really want to check/change.

@WasabiFan


JohnGilb:
I believe the tests.

PR #12, opened 2015-11-07
Add initial version of scripted command infrastructure

WasabiFan:
@aschokking @JohnGilb


src/xbot/common/command/scripted/CommandScriptInterface.java
```diff
@@ -0,0 +1,85 @@
...
+                    log.debug("Waiting indefinitely for lock notification");
+                    lock.wait();
+                }
+                else {
+                    log.debug("Waiting with a timeout of " + timeoutMillis + "ms");
+                    lock.wait(timeoutMillis);
+                }
+            }
+            
+            log.debug("Wait completed (waited " + (System.currentTimeMillis() - startTime) + "millis)");
```
aschokking: I feel like the callers might want to know if it completed successfully or timed out?

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+    
+    /**
+     * For internal use only!
+     * @param command
+     */
+    public synchronized void invokeCommand(Command command) {
+        log.info("Adding command " + command + " to scheduler");
+        
+        Scheduler.getInstance().add(command);
```
aschokking: Why not get the xscheduler by injection instead?

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+            if(!locksToRemove.isEmpty()) {
+                log.debug("Removing " + locksToRemove.size() + " locks");
+                completionLocks.removeAll(locksToRemove);
+            }
+        }
+    }
+
+    @Override
+    protected boolean isFinished() {
+        log.debug("isFinished called; State is " + (execThread == null ? "[uninitialized]" : execThread.getState().toString()));
```
aschokking: Maybe a little cleaner to pull this logic out into a variable so if it ever changes it's not in two places?

tests/xbot/common/command/scripted/ScriptedCommandTest.java
```diff
@@ -0,0 +1,118 @@
...
+        
+        scheduler.add(scriptedCommand);
+        
+        while(!scriptedCommand.isFinished()) {
+            scheduler.run();
+            sleepThread(loopWaitIncrement);
+        }
+        
+        // We need to make sure that it executes at least the limit # of times
+        for(int i = 0; i < numExecs + 1; i++) {
```
aschokking: I'm confused why you need this in addition to the while loop above? Shouldn't it cover doing enough loops?

tests/xbot/common/command/scripted/TestCommandFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import xbot.common.command.scripted.ScriptedCommandFactory;
+import xbot.common.command.scripted.ScriptedCommandProvider;
+
+@Ignore
+public class TestCommandFactory implements ScriptedCommandFactory {
+    private ExecutionCounterCommandProvider lastExecutionCounterCommandProvider;
+    
+    @Override
+    public ScriptedCommandProvider getProviderForName(String commandTypeName) {
+        if(commandTypeName.equals("CounterCommand"))
```
aschokking: Are we going to have to add an entry here for every command to be used in the script?

src/xbot/common/command/scripted/CommandScriptInterface.java
WasabiFan: There isn't really a way to do that (as far as I can tell); the `lock.wait` method doesn't return anything indicating what happened. I could compare my timer value to the timeout value that I passed to `lock.wait`, but that is prone to errors and isn't an option for reliable results.

src/xbot/common/command/scripted/ScriptedCommand.java
WasabiFan: Injection in this case is more code for no extra functionality, and I was pretty lazy :wink:... I'll change this to make it uniform *JIC*.

WasabiFan: Actually, that's more difficult than I had thought; I'd need to remove the constructor params and instead use a separate method. I think it's cleaner as it is now, but obviously I'm open to input -- is there a separate injection construct that I should be using here?

src/xbot/common/command/scripted/ScriptedCommand.java
WasabiFan: Where's the other instance that I have that logic? Or are you saying that I should move that `toString` logic to a separate method in case I decide I want to log the state somewhere else?

src/xbot/common/command/scripted/CommandScriptInterface.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class CommandScriptInterface {
+    static Logger log = Logger.getLogger(CommandScriptInterface.class);
+    
+    private Command wrappedCommand;
+    private ScriptedCommand parentCommand;
+
+    public CommandScriptInterface(Command wrappedCommand, ScriptedCommand parentCommand) {
+        this.wrappedCommand = wrappedCommand;
+        this.parentCommand = parentCommand;
```
JohnGilb: go ahead and add a "init" logging here - it's really nice to see in the logs to see what's getting initialized.

src/xbot/common/command/scripted/CommandScriptInterface.java
```diff
@@ -0,0 +1,79 @@
+package xbot.common.command.scripted;
+
+import org.apache.log4j.Logger;
+
+import edu.wpi.first.wpilibj.command.Command;
```
JohnGilb: Let's only support BaseCommand.

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+    private Set<Command> invokedCommands;
+    
+    private volatile Set<ScriptedCommandCompletionLock> completionLocks;
+    
+    public ScriptedCommand(File scriptFile, ScriptedCommandFactory availableCommandFactory) {
+        this.availableCommandFactory = availableCommandFactory;
+        
+        this.scriptFile = scriptFile;
+        
+        log.debug("Instantiated with file path\"" + scriptFile.getAbsolutePath() + "\"");
```
JohnGilb: what is scriptFile is null, or not a file?

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+
+import edu.wpi.first.wpilibj.command.Command;
+import edu.wpi.first.wpilibj.command.Scheduler;
+
+/**
+ * A command to facilitate execution of a JavaScript script with access
+ * to robot functionality such as Command invocation.
+ *
+ */
+public class ScriptedCommand extends Command {
```
JohnGilb: BaseCommand!

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+        this.availableCommandFactory = availableCommandFactory;
+        
+        this.manualScriptText = manualScriptText;
+        this.manualScriptName = manualScriptName;
+        
+        log.debug("Instantiated with script string");
+    }
+    
+    private void initializeExecThread() {
+        log.debug("Initializing exec thread");
```
JohnGilb: It looks like this is only invoked once - might be a good candidate for info logging.

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    protected void initialize() {
+        initializeExecThread();
+    }
+
+    @Override
+    protected void execute() {
+        // Start the execution thread if this is the first command run
+        if(execThread.getState() == State.NEW) {
+            log.debug("Starting exec thread");
```
JohnGilb: Good info level, I think

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Override
+    protected boolean isFinished() {
+        log.debug("isFinished called; State is " + (execThread == null ? "[uninitialized]" : execThread.getState().toString())); 
+        return execThread != null && execThread.getState() == State.TERMINATED;
+    }
+
+    @Override
+    protected void end() {
+        log.debug("end called");
```
JohnGilb: Anything in the one-time call methods, like interrupted & end, can probably be info logging, since it won't spam the logs.

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+        if(execThread != null) {
+            log.debug("Interrupting exec thread");
+            this.execThread.interrupt();
```
JohnGilb: Does this call "interrupted" on the command currently running?

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,164 @@
...
+                command.cancel();
+            }
+            
+            this.invokedCommands = null;
+        }
+    }
+
+    @Override
+    protected void interrupted() {
+        end();
```
JohnGilb: Info log here

tests/xbot/common/command/scripted/ScriptedCommandTest.java
```diff
@@ -0,0 +1,118 @@
+package xbot.common.command.scripted;
+
+import static org.junit.Assert.*;
+
+import org.apache.log4j.Logger;
+import org.junit.Test;
+
+import xbot.common.command.scripted.ScriptedCommand;
+
+public class ScriptedCommandTest extends BaseScriptedCommandTest {
```
JohnGilb: You are missing a few tests:
-You need a test that loads a script from a file - that's how the robot is going to do it, so you want to make sure that path is solid.
-One where the file does not exist
-One while the script calls a command that does not exist
-One where the file/script is empty/null string
-One where your scripted command is interrupted while some work is going on underneath it

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,222 @@
...
+    @Override
+    public void initialize() {
+        log.info("Initialized");
+        initializeExecThread();
+    }
+
+    @Override
+    public void execute() {
+        // Start the execution thread if this is the first command run
+        if(execThread != null && execThread.getState() == State.NEW) {
```
aschokking: why not do this in initialize if it's a first run thing?

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,222 @@
...
+            this.invokedCommands = null;
+        }
+        else {
+            log.info("No invoked commands to cancel");
+        }
+    }
+
+    @Override
+    public void interrupted() {
+        log.info("Interrupted");
```
aschokking: Isn't this the default interrupted behaviour already?

src/xbot/common/command/scripted/ScriptedCommand.java
```diff
@@ -0,0 +1,222 @@
...
+        end();
+    }
+    
+    public boolean hasReachedCheckpoint(String checkpointName) {
+        return execThread != null && execThread.hasReachedCheckpoint(checkpointName);
```
aschokking: Is this just for testing? Maybe add a comment or something about that?

src/xbot/common/command/scripted/functions/RequireCommandsFunction.java
```diff
@@ -0,0 +1,30 @@
+package xbot.common.command.scripted.functions;
+import java.util.Arrays;
+import java.util.function.Consumer;
+
+import org.mozilla.javascript.Context;
+import org.mozilla.javascript.Scriptable;
+
+/**
+ * A JavaScript-callable function to allow scripts to specify that they require access
```
aschokking: Could elaborate a bit more on what this require access thing is about? Conceptually any command you plan to use in the script?

src/xbot/common/command/scripted/ScriptedCommand.java
WasabiFan: @aschokking because it starts the script immediately, so I don't want to start it until exec is called.

src/xbot/common/command/scripted/ScriptedCommand.java
WasabiFan: Uhhh... You mean the logging? It might be -- can check.

src/xbot/common/command/scripted/ScriptedCommand.java
WasabiFan: No; it's a feature -- the idea is that if the script is going to do something in parallel with something else in our system it can signal when certain things happen. Although it _is_ most useful in tests.

JohnGilb:

PR #13, opened 2015-11-10
Add new readme content for (future) people interested in the project

WasabiFan:
In preparation for the eventual promotion of this library wherever we decide to do it, this should add some useful content for prospective users. To look at the formatted version, you can open the branch page itself.

@aschokking @JohnGilb

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 
 ## What is it?
 
-Seriously Common Lib is a collection of common solutions to robot problems that FRC teams face while using WPILib and Java. It attempts to streamline basic functionality, provide the framework for more advanced programs, and promote clean and well-written code. It is mainly a wrapper around WPILib which enables extra custom functionality.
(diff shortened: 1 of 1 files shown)


aschokking:
Looks great Kaelin, nice job!
Suggestions:
- Maybe make the unit testability more of a thing upfront, maybe explain our motivation for wanting it? Eg multiple people working together, late/limited access to actual robot.
- Pull out the re-usable robot logic bits from the utilities (eg drive system, autonomous, tilt detection). I feel like I would try and sell the base framework to teams that might want to write their own drive code and sort of say ours is there as a bonus instead of an expected thing you would need to use.
- Maybe encourage people to contribute back in if they want, we'd be happy to work on it with others?

WasabiFan:
Sounds good. I think we should also start adding javadoc as we go so that we end up with some reasonable documentation, and maintaining a wiki for big ideas and structure. I'd say that the GitHub wiki should be good for that type of public documentation when we get to it; what do you think? 

As a note to myself: I should make a note about new teams vs. more advanced teams; new teams stick to the pre-built stuff (or maybe stay away from the common lib entirely), while advanced teams look at the complex functionality.

aschokking:
Yeah I think reasonable javadoc on at least the public api surface makes a lot of sense.

WasabiFan:
@aschokking Just updated with the suggested modifications.

aschokking:
lgtm!

PR #15, opened 2015-11-14
Feature/ant checkstyle

aschokking:
Enable running checkstyle rules from Ant so travis ci can vet them.

Current setup has zero tolerance in src files and reasonably relaxed in the tests.

@WasabiFan @JohnGilb


WasabiFan:
@aschokking Your checkstyle job is _very_ unhappy with the NavX code... can you make it ignore the `xbot.common.controls.sensors.navx` package/folder?

aschokking:
It was set to ignore it, unfortunately the pathing was a little different
on travis so I had to tweak things a bit, should be green now.

WasabiFan:
The issue is that there's a new IMU codebase - called "NavX" - that was merged since you diverged from master. The build indicator is telling us that your branch runs fine, but the merged version would not -- so you need to add an extra ignore statement to ignore NavX as well as Nav6 to make it pass when merged.

aschokking:
Ahh that explains it, missed the distinction there. We should maybe put
them all into a 'third party' package or something so we don't have to keep
updating the ignore rules.

WasabiFan:
Yeah maybe we should make a 'libs' package? Should be pretty easy.

PR #17, opened 2015-11-21
Add ability to write test-only assertions in common code

WasabiFan:
This lets us both throw exceptions and make assertions using this helper and have it only blow up if it's running in a test.

@JohnGilb @aschokking


src/xbot/common/logging/SafeRobotAssertionManager.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.log4j.Logger;
+
+import com.google.inject.Singleton;
+
+@Singleton
```
aschokking: Is there a reason this needs to be a singleton?

tests/xbot/common/injection/BaseWPITest.java
```diff
@@ -38,6 +39,9 @@ public void setUp() {
         RobotState.SetImplementation(mockRobotState);
 
         propertyManager = injector.getInstance(PropertyManager.class);
+        
+        SafeRobotAssertionManager assertionMan = injector.getInstance(SafeRobotAssertionManager.class);
+        assertionMan.setExceptionsEnabled(true);
```
aschokking: It would be cleaner to have this set as part of the UnitTestModule() at the guice level so people don't need to remember to do this themselves.

aschokking:
lgtm;

tests/xbot/common/injection/BaseWPITest.java
JohnGilb: Agree

src/xbot/common/logging/SafeRobotAssertionException.java
```diff
@@ -0,0 +1,7 @@
+package xbot.common.logging;
+
+public class SafeRobotAssertionException extends RuntimeException {
+    public SafeRobotAssertionException(String faliureCauseCause) {
```
WasabiFan: Note to self: typo

WasabiFan:
@JohnGilb @aschokking Can you re-review?

src/xbot/common/logging/SilentRobotAssertionManager.java
```diff
@@ -0,0 +1,15 @@
+package xbot.common.logging;
+
+public class SilentRobotAssertionManager extends RobotAssertionManager {
+
+    @Override
+    protected void handlePlatformException(RuntimeException e) {
+        // Don't do anything: we don't need to throw
```
JohnGilb: Log a warning! We don't want the robot to die, but we should probably know about this.

PR #18, opened 2015-12-12
Features/IMUTests

robeat101:
Added IMU stuff for both real and mock controllers.


WasabiFan:
Looks good.

We also have a [Gyro class](https://github.com/Team488/SeriouslyCommonLib/blob/master/src/xbot/common/controls/sensors/XGyro.java) which seems very similar (although poorly named) -- these should probably be merged, unless there's a significant difference that I can't discern.

JohnGilb:
@aschokking @WasabiFan Check this out, updated to support NavX.

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
```diff
@@ -98,14 +100,24 @@ public AdvancedJoystickButton getJoystickButton(XJoystick joystick, int button_n
...
     @Override
-    public XGyro getGyro() {
+    public XGyro getGyro(ImuType imuType) {
         // It's possible that the nav6 might get disconnected, and throw some
         // exceptions
         // at runtime when it can't communicate over the serial port.
         // The robot needs to protect itself from this behavior.
         try {
-            return new Nav6Gyro();
+            if (imuType == ImuType.nav6) {
```
aschokking: A little cleaner as a case statement maybe? Or if you want if's, make the subsequent ones else ifs just to be explicit?

JohnGilb: agreed, and changed.

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
```diff
@@ -98,14 +100,23 @@ public AdvancedJoystickButton getJoystickButton(XJoystick joystick, int button_n
...
         // at runtime when it can't communicate over the serial port.
         // The robot needs to protect itself from this behavior.
         try {
-            return new Nav6Gyro();
+            switch (imuType) {
+            case nav6:
+                return new Nav6Gyro();
+            case navX:
+                return new InertialMeasurementUnitAdapter(Port.kMXP);
+                default:
```
aschokking: weird tabbing on default? Also shouldn't call the case statements be tabbed in from switch?

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
```diff
@@ -98,14 +100,23 @@ public AdvancedJoystickButton getJoystickButton(XJoystick joystick, int button_n
...
+                    log.error("Could not find " + imuType.name() + "! Returning a \"broken\" MockGyro instead.");
```
aschokking: This should maybe be a test-time-only exception?

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
JohnGilb: Fixed.

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
JohnGilb: Good idea.

JohnGilb: Actually, I'm not sure. This code can't be hit during tests, since it is in the RealWpiFactory, which is never directly tested due to WPILib.

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+    public InertialMeasurementUnitAdapter() {
+        super(ImuType.navX);
+        /* Options: Port.kMXP, SPI.kMXP, I2C.kMXP or SerialPort.kUSB */
+        try {
+            this.ahrs = new AHRS(Port.kMXP);
+        }
+        catch (Exception e){
+            isBroken = true;
+        }
+        this.ahrs.zeroYaw();
```
WasabiFan: Are we sure that we want to zero the yaw on initialization?

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+public class InertialMeasurementUnitAdapter extends NavImu implements XGyro {
+
+    AHRS ahrs;
+    boolean isBroken = false;
+
+    public InertialMeasurementUnitAdapter() {
+        super(ImuType.navX);
+        /* Options: Port.kMXP, SPI.kMXP, I2C.kMXP or SerialPort.kUSB */
+        try {
+            this.ahrs = new AHRS(Port.kMXP);
```
WasabiFan: I think this should be an (optional?) constructor argument so that we can use this to connect through other ports if needed.

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public double getPitch() {
+        return this.ahrs.getPitch();
+    }
+
+    @Override
+    public boolean isBroken() {
+        // TODO Auto-generated method stub
```
WasabiFan: This is no longer TODO

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
WasabiFan: Looks like you haven't pushed those changes yet, FYI

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
```diff
@@ -114,12 +125,16 @@ public XGyro getGyro() {
...
             log.error("Could not create gyro! Returning a \"broken\" MockGyro instead.");
-
-            MockGyro brokenGyro = new MockGyro(new MockRobotIO());
-            brokenGyro.setIsBroken(true);
-            return brokenGyro;
+            return getBrokenGyro();
         }
     }
+    
+    private XGyro getBrokenGyro()
```
WasabiFan: Can we add an optional argument `isBroken` in the `MockGyro` constructor? This would be much more elegant and easy to understand at a glance as a one-liner.

src/xbot/common/properties/RobotDatabaseStorage.java
```diff
@@ -2,7 +2,7 @@
 
 public class RobotDatabaseStorage extends DatabaseStorageBase {
 
-    private static String location = "/TeamDatabase";
+    private static String location = "/home/lvuser/TeamDatabase";
```
WasabiFan: - This should be `final`
- It feels wrong to hard-code the location of the database like this as an absolute path... not sure how we can fix that though.

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
JohnGilb: Hmm... good point. There are some edge cases where it might be a good idea for the IMU to keep state. It's hard to imagine they'd be hit, but better safe then sorry.

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
JohnGilb: Noted

src/xbot/common/injection/wpi_factories/RealWPIFactory.java
JohnGilb: That seems reasonable.

src/xbot/common/properties/RobotDatabaseStorage.java
JohnGilb: - Good call on final
- I think hardcoding here is OK - that's sort of the point of this injected class.

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
JohnGilb: Fixed.

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
JohnGilb: Fixed

PR #19, opened 2016-01-13
Update JARs in lib/ dir for 2016

WasabiFan:
This replaces the 2015 JARs with the new ones. Not sure if we want to wait to merge this until we test it on-robot.

It would be a good idea to tag a release on master before we merge this so we have an easy checkpoint. I can do this if that's what we want.

@JohnGilb @aschokking


aschokking:
@WasabiFan I think your plan of waiting for the on-robot test is a good one, we should keep master in a known good config whenever possible. Adding a release tag sounds good too, go for it!

WasabiFan:
@aschokking I just created a release draft [here](https://github.com/Team488/SeriouslyCommonLib/releases/tag/untagged-04e0ebd023aa932d43dd). I used the version number `v0.1`, because it seems like a good initial version. Can you look at it and then hit the publish button if it looks good?

WasabiFan:
This PR has been updated to also make the build scripts less environment-dependent.

aschokking:
lgtm

PR #22, opened 2016-01-30
Stop using database for most unit tests

aschokking:
Add a level of abstraction infront of the database. Only use the real database for those tests but use in memory table for everything else.
@JohnGilb @WasabiFan

### DIFF
--- a/src/xbot/common/properties/PermanentStorage.java
+++ b/src/xbot/common/properties/PermanentStorage.java
@@ -1,0 +1,5 @@
+package xbot.common.properties;
+
+public interface PermanentStorage extends ITableProxy {
+
+}
--- a/tests/xbot/common/injection/BaseWPITest.java
+++ b/tests/xbot/common/injection/BaseWPITest.java
@@ -28,11 +28,11 @@ public class BaseWPITest {
 
     public PropertyManager propertyManager;
     
-    protected AbstractModule guice_module = new UnitTestModule();
+    protected AbstractModule guiceModule = new UnitTestModule();
 
     @Before
     public void setUp() {
-        injector = Guice.createInjector(guice_module);
+        injector = Guice.createInjector(guiceModule);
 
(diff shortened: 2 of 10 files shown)


src/xbot/common/properties/PermanentStorage.java
```diff
@@ -0,0 +1,5 @@
+package xbot.common.properties;
+
+public interface PermanentStorage extends ITableProxy {
```
WasabiFan: Why do we have a plank interface?

tests/xbot/common/injection/BaseWPITest.java
```diff
@@ -26,10 +27,12 @@
     public MockRobotState mockRobotState;
 
     public PropertyManager propertyManager;
+    
+    protected AbstractModule guice_module = new UnitTestModule();
```
WasabiFan: Should be camelCase

tests/xbot/common/injection/UnitTestModule.java
```diff
@@ -27,7 +32,13 @@ protected void configure() {
         this.bind(WPIFactory.class).to(MockWPIFactory.class);
 
         this.bind(ITableProxy.class).to(TableProxy.class).in(Singleton.class);
-        this.bind(DatabaseStorageBase.class).to(OffRobotDatabaseStorage.class).in(Singleton.class);
+        
+        if(this.useRealDatabaseForPropertyStorage) {
```
WasabiFan: We should use a variable here to store the target class instead of duplicating the rest of the line

src/xbot/common/properties/PermanentStorage.java
aschokking: The blank interface is to allow setting a particular implementation in guice land, it just happens to be at this point we don't need anything in addition to ITableProxy right now but if we do need something for storage it can go here.

PR #23, opened 2016-02-06
Have the JoystickButtonManager throw exceptions during tests and retu…

JohnGilb:
…rn mock buttons during runtime (if somebody tries to get an illegal button)


tests/xbot/common/controls/sensors/JoystickButtonManagerTest.java
```diff
@@ -16,25 +19,36 @@ public void testAvailability() {
...
+            testButton(testButtons, x);
         }
 
     }
+    
+    private void testButton(JoystickButtonManager manager, int i) {
+        try {
+            manager.getifAvailable(i);
+            assertTrue(false);
+        } catch (Exception e)
```
WasabiFan: Newline after the close brace

tests/xbot/common/controls/sensors/JoystickButtonManagerTest.java
```diff
@@ -16,25 +19,36 @@ public void testAvailability() {
...
-            assertTrue("Button " + x + " should be null.", null == testButtons.getifAvailable(x));
+            testButton(testButtons, x);
         }
 
     }
+    
+    private void testButton(JoystickButtonManager manager, int i) {
+        try {
+            manager.getifAvailable(i);
+            assertTrue(false);
```
WasabiFan: You should be able to `fail()`

src/xbot/common/controls/sensors/JoystickButtonManager.java
```diff
@@ -44,14 +52,28 @@ public void addAnalogButton(AnalogHIDDescription desc) {
...
         } else {
             // Warn people that terrible things are happening, then return a null button.
-            log.error("button " + buttonNumber + " is already used! Cannot be used twice!");
-            return null;
+            String message = "button " + buttonNumber + " is already used! Cannot be used twice!";
+            return getMockButton(message);
         }
     }
+    
+    private AdvancedJoystickButton getMockButton(String message) {
```
WasabiFan: I don't think that we want a method called `getMockButton` to log as well and throw exceptions. I think it would be better to either:
- Log and throw outside of the method instead
- Rename it to `handleBadButton` (or similar)

src/xbot/common/controls/sensors/JoystickButtonManager.java
```diff
@@ -44,14 +52,28 @@ public void addAnalogButton(AnalogHIDDescription desc) {
     }
 
     public AdvancedJoystickButton getifAvailable(int buttonNumber) {
+        
+        if (buttonNumber < 1 || buttonNumber > maxButtons) {
+            String message = "button " + buttonNumber + " is out of range!";
```
WasabiFan: I think we can streamline this to one line -- it seems like it would be easier to read.

tests/xbot/common/controls/sensors/JoystickButtonManagerTest.java
```diff
@@ -16,25 +19,36 @@ public void testAvailability() {
...
             assertTrue("Button " + x + " should not be null.", null != testButtons.getifAvailable(x));
         }
         for (int x = 1; x <= 12; x++) {
-            assertTrue("Button " + x + " should be null.", null == testButtons.getifAvailable(x));
+            testButton(testButtons, x);
         }
 
     }
+    
+    private void testButton(JoystickButtonManager manager, int i) {
```
WasabiFan: jUnit has an annotation for expected exceptions. I think it would be best to actually split this up into separate tests and set the exception that you want to see.

PR #29, opened 2016-02-21
Add output limits to pid manager

aschokking:
defaults to -1 1

### DIFF
--- a/src/xbot/common/math/PIDManager.java
+++ b/src/xbot/common/math/PIDManager.java
@@ -13,8 +13,8 @@ public class PIDManager {
     private DoubleProperty propP;
     private DoubleProperty propI;
     private DoubleProperty propD;
-    private DoubleProperty maxPositiveOutput;
-    private DoubleProperty minNegativeOutput;
+    private DoubleProperty maxOutput;
+    private DoubleProperty minOutput;
 
     public PIDManager(String functionName, XPropertyManager propMan, double defaultP, double defaultI, double defaultD,
             double defaultMaxOutput, double defaultMinOutput) {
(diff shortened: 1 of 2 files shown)


aschokking:
@JohnGilb

src/xbot/common/math/PIDManager.java
```diff
@@ -13,26 +13,37 @@
...
-            double defaultD) {
-        propP = new DoubleProperty(functionName + " P", defaultP, propMan);
-        propI = new DoubleProperty(functionName + " I", defaultI, propMan);
-        propD = new DoubleProperty(functionName + " D", defaultD, propMan);
+            double defaultD, double defaultMaxOutput, double defaultMinOutput) {
+        propP = propMan.createPersistentProperty(functionName + " P", defaultP);
+        propI = propMan.createPersistentProperty(functionName + " I", defaultI);
+        propD = propMan.createPersistentProperty(functionName + " D", defaultD);
+        
+        maxOutput = propMan.createPersistentProperty(functionName + " MaxOutput", defaultMaxOutput);
```
WasabiFan: There should be spaces between words

src/xbot/common/math/PIDManager.java
```diff
@@ -13,26 +13,37 @@
...
-        setupPIDManager(functionName, propMan, defaultP, defaultI, defaultD);
+        setupPIDManager(functionName, propMan, defaultP, defaultI, defaultD, 1.0, -1.0);
     }
 
     public PIDManager(String functionName, XPropertyManager propMan) {
-        setupPIDManager(functionName, propMan, 0, 0, 0);
+        setupPIDManager(functionName, propMan, 0, 0, 0, 1.0, -1.0);
     }
 
     private void setupPIDManager(String functionName, XPropertyManager propMan, double defaultP, double defaultI,
```
WasabiFan: Why do we have this method? I think this should go in the most verbose constructor and then the others should just cascade down from least to most explicit.

JohnGilb:
lgtm

src/xbot/common/math/PIDManager.java
aschokking: I'm not sure we ever really settled on a philosophy for what things should look like in the property names. I feel like most of the robot ones are camel cased for instance.

src/xbot/common/math/PIDManager.java
aschokking: Sounds reasonable, not sure why it was done this way originally

src/xbot/common/math/PIDManager.java
WasabiFan: There isn't any technical reason to omit spaces (and you include them to separate the value from the label already). So, given that humans are better-able to read space-separated words, we should target the most readable option.

PR #30, opened 2016-02-21
Encoders now automatically have Distance Per Pulse as a property.

JohnGilb:


src/xbot/common/controls/sensors/wpi_adapters/EncoderWPIAdapter.java
```diff
@@ -2,30 +2,28 @@
...
 public class EncoderWPIAdapter implements XEncoder {
 
     Encoder adapter;
     boolean inverted;
+    public DoubleProperty distancePerPulse;
 
-    public EncoderWPIAdapter(int aChannel, int bChannel) {
+    public EncoderWPIAdapter(String name, int aChannel, int bChannel, double defaultDistancePerPulse, XPropertyManager propMan) {
         adapter = new Encoder(aChannel, bChannel);
+        distancePerPulse = propMan.createPersistentProperty(name + "-DistancePerPulse", defaultDistancePerPulse);
```
WasabiFan: Spaces! Put spaces in property names!

JohnGilb: rather than pretty dashes? =] @WasabiFan

JohnGilb: camel casing here seems appropriate overall though - space on the dashboard is at a slight premium, especially with how many properties we have.

WasabiFan: I see the point on space availability, but I think that readability is a very important factor to consider; humans read space-separated words much more easily than a single block of characters, even if each word is capitalized. Prioritizing the 1/8 space savings over the ability to glance at properties to read then seems like a pretty bad tradeoff.

PR #31, opened 2016-02-24
Use real value types instead of references in property manager

WasabiFan:
Also cleaned up some pointless comments in the file.

@aschokking @JohnGilb


JohnGilb:
This looks good

PR #34, opened 2016-03-19
Fix NavX/MXP pinout mappings

WasabiFan:


tests/xbot/common/injection/wpi_factories/MockWPIFactory.java
```diff
@@ -74,6 +74,10 @@ private void checkAnalog(int channel) {
     
     private void checkDio(int channel) {
         if (channel >= 10) {
+            if(channel > 13 && channel < 18) {
+                throw new RuntimeException("Allocated MXP digital pin that does not exist!");
```
aschokking: Might be nice to list valid dio pins out in the message?

PR #35, opened 2016-08-17
Add CAN talon interface

WasabiFan:
This is an initial implementation of the CAN talon's interface. I omitted some niche features and some duplicated methods to keep the MVP as small as is reasonable, but there are still over 100 methods in this interface. For now, the raw voltage output is exposed as PWM channels, where the talon's device ID is negated and used as the PWM channel, but I'd like to move away from that if we can get the other testing infrastructure modified to support a different addressing mechanism. There are some interfaces (such as the fault/sticky fault getters) that I didn't fully mock up.

I was considering adding tests of the mock implementation, but for now at least I'm not going to because I need to focus on the actual robot code. @aschokking @JohnGilb If you have the time, a sanity review before Saturday would be appreciated; there's a lot of surface area for bugs to pop up

The interfaces and methods that I omitted are:
  - `enableControl`/`disableControl` (same as `enable`/`disable`)
  - `delete`
  - motion profile control mode
  - "expiration" and other MotorSafety members
  - `getPinStateQuad*`
  - `getPulseWidth*`
  - `setPID` w/ special args
  - `setParameter`/`getParameter`

Also, I noticed that the `MockEncoder` takes arguments for the channels, but the arguments are never used. I added another constructor without arguments, but we might just want to remove the constructor altogether.


WasabiFan:
I can't tell why the CI compile error is occurring. Once that is cleared up, checkstyle is also complaining about naming, so I need to figure out how to override that.

WasabiFan:
I also renamed some of the methods from the official interface to make them less stupid, e.g. `rev` -> `reverse` and `for` -> `forward`.

WasabiFan:
CI fixed.

WasabiFan:
A review of my progress on the main robot control code which uses this feature would also be appreciated, but is less important: https://github.com/Team488/summer-2016-robot/compare/ca0b9b74ca6c91373ea06803486e6ea58906290a...master

aschokking:
Man that's a lot of methods, nice job!

JohnGilb:
Overall good stuff.

If you have time before this gets deployed, I'd suggest even a simple sanity check of a test case. (Get a motor, drive it forward full power in a simple mode, verify a positive power in the mockIO).

PR #37, opened 2016-09-17
Feature/setpoint systems

JohnGilb:
Adds the abstract classes BaseSetpointCommand and BaseSetpointSystem.

The idea: for a lot of maintainer/setpoint systems, we have been creating an extra "dummy" subsystem to use as a lock for all the setpoint commands.

With this new approach, if you want your Subsystem to be "setpoint-able", you extend it from BaseSetpointSubsystem, which automatically creates a dummy subsystem for use by BaseSetpointCommands (which automatically require the dummy subsystem without any code from you).

### DIFF
--- a/src/xbot/common/command/BaseSetpointCommand.java
+++ b/src/xbot/common/command/BaseSetpointCommand.java
@@ -1,8 +1,8 @@
 package xbot.common.command;
 
-public abstract class BaseSetpointCommand extends BaseCommand{
+public abstract class BaseSetpointCommand extends BaseCommand {
 
-    public BaseSetpointCommand(SupportsSetpoint system) {
-        requires(system.getSetpointSystem());
+    public BaseSetpointCommand(SupportsSetpointLock system) {
+        requires(system.getSetpointLock());
     }
 }
--- a/src/xbot/common/command/BaseSetpointCommand.java
+++ b/src/xbot/common/command/BaseSetpointCommand.java
@@ -1,0 +1,8 @@
+package xbot.common.command;
+
+public abstract class BaseSetpointCommand extends BaseCommand{
+
+    public BaseSetpointCommand(SupportsSetpoint system) {
+        requires(system.getSetpointSystem());
+    }
+}
--- a/src/xbot/common/command/SupportsSetpoint.java
+++ b/src/xbot/common/command/SupportsSetpoint.java
@@ -1,0 +1,8 @@
+package xbot.common.command;
+
+import edu.wpi.first.wpilibj.command.Subsystem;
+
+public interface SupportsSetpoint {
+
+    public Subsystem getSetpointSystem();
+}
--- a/tests/xbot/common/command/SetpointSystemTest.java
+++ b/tests/xbot/common/command/SetpointSystemTest.java
@@ -1,0 +1,46 @@
+package xbot.common.command;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import edu.wpi.first.wpilibj.command.Scheduler;
+import xbot.common.injection.BaseWPITest;
+
+public class SetpointSystemTest extends BaseWPITest{
+
+    @Before
+    public void setUp() {
+        super.setUp();
+    }
+    
+    @Test
+    public void testSetpointSystemCanBeCreated() {
+        MockSetpointCommand first = injector.getInstance(MockSetpointCommand.class);
+    }
+    
+    @Test
+    public void testSetpointCommandsCollide() {
+        MockSetpointCommand first = injector.getInstance(MockSetpointCommand.class);
+        MockSetpointCommand second = injector.getInstance(MockSetpointCommand.class);
+        
+        XScheduler xScheduler = this.injector.getInstance(XScheduler.class);
+        
+        assertFalse("First command is not running", first.isRunning());
+        assertFalse("Second command is not running", second.isRunning());
+        
+        Scheduler.getInstance().add(first);
+        xScheduler.run();
+        assertTrue("First command is running", first.isRunning());
+        
+        Scheduler.getInstance().add(second);
+        xScheduler.run();
+        
+        assertTrue("Second command is running", second.isRunning());
+        assertFalse("First command is no longer running", first.isRunning());
+    }
+    
+    
+}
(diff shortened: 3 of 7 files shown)


src/xbot/common/command/BaseSetpointCommand.java
```diff
@@ -0,0 +1,8 @@
+package xbot.common.command;
+
+public abstract class BaseSetpointCommand extends BaseCommand{
```
aschokking: space after BaseCommand

src/xbot/common/command/SupportsSetpoint.java
```diff
@@ -0,0 +1,8 @@
+package xbot.common.command;
+
+import edu.wpi.first.wpilibj.command.Subsystem;
+
+public interface SupportsSetpoint {
```
aschokking: this is more specific than maybe SupportsSetPoint name implies, maybe `SupportsSetpointSubsystem`? Like I know wpi has it's own similar interface

tests/xbot/common/command/SetpointSystemTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Test;
+
+import edu.wpi.first.wpilibj.command.Scheduler;
+import xbot.common.injection.BaseWPITest;
+
+public class SetpointSystemTest extends BaseWPITest{
+
+    @Before
+    public void setUp() {
+        super.setUp();
```
aschokking: why have a setup?

src/xbot/common/command/BaseSetpointCommand.java
aschokking: testing sub comment

tests/xbot/common/command/SetpointSystemTest.java
JohnGilb: Thought I needed it, or it wouldn't call it on the base class - but just tested it now and it works! So, removing it.

src/xbot/common/command/BaseSetpointCommand.java
JohnGilb: Thanks

src/xbot/common/command/SupportsSetpoint.java
JohnGilb: Renamed

WasabiFan:
In the future, I think we should try to use the squash option for merging PRs, especially when there are "respond to feedback" commits.

JohnGilb:
@WasabiFan you can see a trivial example in the tests, with MockSetpointSystem and MockSetpointCommand.

Basic idea: For any subsystem that is going to be "maintained" by a command, have it extend from BaseSetpointSubsystem. For any commands that are going to set a setpoint on that subsystem, have them extend from BaseSetpointCommand.

This will automatically take care of the requires() aspect for you.

WasabiFan:
Ok... What's the function of the `SupportsSetpointLock` interface? What does it provide that locking on the subsystem itself does not?

JohnGilb:
Just a way of abstracting out what exactly is providing the lock object -
as far as the BaseSetpointCommand is concerned, it doesn't necessarily have
to be a subsystem, just some object that implements the
SupportsSetpointLock interface.

Right now, the only object that does that is BaseSetpointSubsystem, but
there could be others in the future.

WasabiFan:
Is it intended that the lock on the subsystem and the lock on the "lockable" object are independent? Would that not introduce the potential for confusion?

JohnGilb:
The idea is that a subsystem (which is the lock object between manual control / maintainer commands) contains another subsystem inside of it (which is the lock object for setpoint commands)

WasabiFan:
It seems like we are attempting to use subsystems to model something that subsystems really aren't designed to do, and as a result we are going to be building convoluted systems which don't interact in a natural way. I still don't fully understand how this set of classes applies to code in a way that makes it simpler, and as a result I can't make any real judgement on whether I think it is actually a good solution or not. Having said that, I suspect the fundamental issue is that it is attempting to solve this problem at the wrong level.

What happens if there are multiple types of setpoint, for example? e.g. The turret rotation is sometimes controlled by velocity and other times by position. When transitioning from position to velocity, there is logic that needs to run to manage the transition.

JohnGilb:
This code was added to streamline an existing robot design pattern - the Subsystem and "SetpointSubsystem" pair.

For systems like the Drive, one manual command and several "maintainer" commands would lock on the DriveSubsystem. DriveWithJoysticksCommand, MaintainPositionPIDCommand, MaintainVelocityPIDCommand, etc...

However, we would also want to control setpoints for the systems. We couldn't have commands require the core subsystem, as this would interrupt the maintainers. Usually, our solution was either:
1) Have the command not require the system - works for simple things like "set forklift to height level 4"
2) (for something dynamically setting the setpoint often) create a setpoint subsystem, which does not directly own any motors or sensors,, and require that.

However, since the setpoint subsystem is so tied to it's "core" subsystem, I decided to streamline that a little by making it a child object inside of the core subsystem, and having SetpointCommands automatically grab that child object as a lock.

This approach does break down if your system needs to have more than one setpoint set at a time - for example, if you have two separate PIDs running simultaneously & independently on a subsystem. However, since most of our subsystems only control one "entity", this shouldn't happen very often.

JohnGilb:
As for multiple types of setpoint, this approach should work fine - the whole point of subsystems is to handle conflict between multiple commands. If the core subsystem is running "MaintainPositionPIDCommand" and the setpoint subsystem is running "GetPositionSetpointFromJoysticks", you just interrupt both of them with "MaintainVelocityPIDCommand" and "GetVelocitySetpointFromFileCommand".

WasabiFan:
OK, I'm with you so far. Am I correct that the benefit that this provides is that inheriting from the new class defines an empty subsystem for use as a lock so that you don't have to define one yourself? Is there anything else that it does?

Looking at this code, I see some points of potential confusion that indicate to me that there is something wrong with the design as a whole:
- The purpose of most of this code is unclear
- There is a completely empty `Subsystem` object
- Most of the interfaces/classes are very small (including one empty class)
- Two classes which a consumer defines must inherit from their respective parents and then remember to call the `super` constructor, otherwise the system will behave unexpectedly (read: explode)
- There are multiple places with potential for `NullPointerException`s

The most worrying items are the first two. We are using these objects for purposes for which they were not intended, and as a result we must maintain this significant amount of boilerplate to work around design pitfalls. I think that a much more natural (not to mention maintainable) solution would be to rework the command/subsystem/scheduler interaction as we've been discussing for years. Short of that, I don't see any way to make this clearer to a reader, and right now the system that we have here is only marginally simpler than implementing the separate subsystem. I like the fact that this structures the involved classes a bit, but it really isn't enough of a benefit to justify the cost as I see it.

JohnGilb:
That's all it does. It takes a pattern we already have and makes it easier to do it again.

Was that pattern the best possible pattern to begin with? Probably not - it was born out of the limitations of our current framework. As a result, once we get around to modifying the scheduler in a serious way (which you allude to above) we'll almost certainly want to get rid of this.

PR #41, opened 2016-12-03
Buttons should start at 1, since the other joysticks assume this

JohnGilb:


src/xbot/common/controls/sensors/wpi_adapters/GamepadJoystickWpiAdapter.java
```diff
@@ -39,10 +39,10 @@
         DPad
     }
     
-    public static int DPadUp = 0;
-    public static int DPadDown = 1;
-    public static int DPadLeft = 2;
-    public static int DPadRight = 3;
+    public static int DPadUp = 1;
```
WasabiFan: These should be constant; as it is, I could redefine this at runtime.

PR #44, opened 2017-01-25
Adds a method to XCANTalon allowing us to ensure that the talon is in…

JohnGilb:
… the control mode we want it to be in. Given that many of our subsystems will use talons in at least two modes (PID and PercentVbus), this should prove useful.

### DIFF
--- a/src/xbot/common/controls/actuators/XCANTalon.java
+++ b/src/xbot/common/controls/actuators/XCANTalon.java
@@ -1,6 +1,7 @@
 package xbot.common.controls.actuators;
 
 import com.ctre.CANTalon;
+import com.ctre.CANTalon.TalonControlMode;
 
 import xbot.common.properties.XPropertyManager;
 
@@ -129,4 +130,5 @@ public interface XCANTalon extends XSpeedController {
     // Custom helpers -------------------------------------
     void createTelemetryProperties(String deviceName, XPropertyManager propertyManager);
     void updateTelemetryProperties();
+    void ensureModeForTalon(TalonControlMode mode);
 }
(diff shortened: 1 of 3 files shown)


src/xbot/common/controls/actuators/XCANTalon.java
```diff
@@ -129,4 +130,5 @@
     // Custom helpers -------------------------------------
     void createTelemetryProperties(String deviceName, XPropertyManager propertyManager);
     void updateTelemetryProperties();
+    void ensureModeForTalon(TalonControlMode mode);
```
WasabiFan: Given that this is in the `XCANTalon` class, I think we can drop the `ForTalon`.

src/xbot/common/controls/actuators/wpi_adapters/CANTalonWPIAdapter.java
```diff
@@ -477,4 +477,11 @@ public void updateTelemetryProperties() {
         temperatureProperty.set(this.getTemperature());
     }
 
+    @Override
+    public void ensureModeForTalon(TalonControlMode mode) {
+        if (this.getControlMode() != mode) {
+            this.setControlMode(mode);
```
WasabiFan: The fact that the two implementations are so different seems odd. What if we made `XCANTalon` into an abstract class?

src/xbot/common/controls/actuators/XCANTalon.java
JohnGilb: Good point! Definitely makes the method name easier to understand.

aschokking:
lgtm

PR #46, opened 2017-01-30
updateTelemetry on BaseSubsystem

aschokking:
* Add an optional `updateTelemetry` method on BaseSubsystem
* Add ability for these subsystems to be registered with BaseRobot class and they will be called automatically during every periodic loop.

The goal here was to make it as low overhead as possible, subsystems just need to override the method and then get registered (so no need for additional boilerplate commands to call updateTelemetry).

### DIFF
--- a/src/xbot/common/command/BaseRobot.java
+++ b/src/xbot/common/command/BaseRobot.java
@@ -1,6 +1,6 @@
 package xbot.common.command;
 
-import java.util.Vector;
+import java.util.ArrayList;
 
 import org.apache.log4j.Logger;
 import org.apache.log4j.xml.DOMConfigurator;
@@ -37,7 +37,7 @@ public class BaseRobot extends IterativeRobot {
     
     protected Command autonomousCommand;
     
-    protected Vector<BaseSubsystem> telemetryUpdatingSubsystems = new Vector<BaseSubsystem>();
+    protected ArrayList<TelemetrySource> telemetrySources = new ArrayList<TelemetrySource>();
 
     public BaseRobot() {
         super();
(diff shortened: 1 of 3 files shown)


src/xbot/common/command/BaseRobot.java
```diff
@@ -34,6 +36,8 @@
     protected Injector injector;
     
     protected Command autonomousCommand;
+    
+    protected Vector<BaseSubsystem> telemetryUpdatingSubsystems = new Vector<BaseSubsystem>();
```
WasabiFan: Is there a reason that this is a `Vector`? They were deprecated in Java 2 IIRC. It seems like this should be a normal `ArrayList`.

src/xbot/common/command/BaseSubsystem.java
```diff
@@ -13,5 +14,10 @@ public void setDefaultCommand(Command command) {
     protected void initDefaultCommand() {
 
     }
+    
+    public void updateTelemetry() {
```
WasabiFan: This architecture feels odd. Not all subsystems have telemetry to report and some things that _aren't_ subsystems may want to supply telemetry, so putting it on the subsystem doesn't make much sense to me. Could we define a `TelemetrySource`, `ScheduledTelemetrySource` or similar interface that has this method on it and then implement that on the subsystems that want to be given telemetry cycles? Then the `BaseRobot` would operate on generic `TelemetrySource`s (which could be, for example, my vision server thread).

src/xbot/common/command/BaseRobot.java
```diff
@@ -118,4 +122,19 @@ public void teleopPeriodic() {
     public void testPeriodic() {
         LiveWindow.run();
     }
+    
+    protected void sharedPeriodic(){
```
WasabiFan: Space before brace

src/xbot/common/command/BaseRobot.java
aschokking: I'm just old :D

src/xbot/common/command/BaseSubsystem.java
aschokking: That's fair, I was debating between the two options and had leaned this way to reduce the number of things people had to learn about (like another interface they need to add to their subsystem). That said, the non-subsystem class case seems like a reasonable one.

src/xbot/common/command/BaseSubsystem.java
```diff
@@ -2,6 +2,7 @@
 
 import edu.wpi.first.wpilibj.command.Command;
 import edu.wpi.first.wpilibj.command.Subsystem;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
WasabiFan: ಠ_ಠ

PR #47, opened 2017-02-04
Feature/pid property manager lite

JohnGilb:
With the use of CANTalons, we are having situations where subsystems like the SideShooter need to create PIDF properties (so we can tune easily via dashboard). This is probably going to happen more than once, so I've created a simple property container class for PIDF values.

I was also thinking about directly integrating this into the XCANTalon, but haven't thought of a good way to do that yet.

### DIFF
--- a/src/xbot/common/math/PIDPropertyManager.java
+++ b/src/xbot/common/math/PIDPropertyManager.java
@@ -40,7 +40,7 @@ public class PIDPropertyManager {
         return propD.get();
     }
 
-    public void setDefaultD(double d) {
+    public void setD(double d) {
         propD.set(d);
     }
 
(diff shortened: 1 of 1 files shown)


WasabiFan:
We should figure out how this coexists with existing classes; my inclination would be to make the existing `PIDManager` extend this one.

src/xbot/common/math/PIDPropertyManager.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public void setI(double i) {
+        propI.set(i);
+    }
+
+    public double getD() {
+        return propD.get();
+    }
+
+    public void setDefaultD(double d) {
```
aschokking: remove Default?

JohnGilb:
Asana task added for PIDManager integration: https://app.asana.com/0/38541457243752/262698628840807

PR #48, opened 2017-02-04
Port the pose subsystem from 2016 and add it to the commonlib

JohnGilb:


src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+        // update the "old" value
+        lastImuHeading = imu.getYaw();
+        
+        currentHeadingProp.set(currentHeading.getValue());
+        
+        currentPitch.set(getRobotPitch());
+        currentRoll.set(getRobotRoll());
+    }  
+    
+    private void updateDistanceTraveled(double currentLeftDistance, double currentRightDistance) {
```
aschokking: might want to mention units inches instead of ticks, maybe in a comment if nothing else?

aschokking: ah nvm, just saw this was private and the public guy below has the units

src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+import xbot.common.controls.sensors.XGyro;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.math.ContiguousHeading;
+import xbot.common.math.XYPair;
+import xbot.common.properties.BooleanProperty;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+@Singleton
+public class PoseSubsystem extends BaseSubsystem {
```
aschokking: this should probably be a TelemetrySource

JohnGilb: After an offline discussion, we decided that since the regular update requires distance inputs, that any "telemetry updates" will happen in the DriveSubsystem (which owns the distance sensors)

src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+    private double previousRightDistance;
+    
+    private BooleanProperty rioRotated;
+    
+    @Inject
+    public PoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
+        log.info("Creating PoseSubsystem");
+        imu = factory.getGyro(ImuType.navX);
+        
+        currentHeadingProp = propManager.createEphemeralProperty("CurrentHeading", 0.0);
```
WasabiFan: Properties should have spaces and proper caps.

src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+        // Right when the system is initialized, we need to have the old value be
+        // the same as the current value, to avoid any sudden changes later
+        lastImuHeading = imu.getYaw();
```
WasabiFan: In the constructor, we really don't want to be accessing any sensors. This should be done in some other way.

src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+        double deltaLeft = currentLeftDistance - previousLeftDistance;
+        double deltaRight = currentRightDistance - previousRightDistance;
+        
+        double totalDistance = (deltaLeft + deltaRight) / 2;
+        
+        // get X and Y
+        double deltaY = Math.sin(currentHeading.getValue() * Math.PI / 180) * totalDistance;
+        double deltaX = Math.cos(currentHeading.getValue() * Math.PI / 180) * totalDistance;
+        
+        totalDistanceX.set(totalDistanceX.get() + deltaX);
```
WasabiFan: We don't really want to be using properties like this.

src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+    
+    private XYPair getTravelVector() {
+        return new XYPair(totalDistanceX.get(), totalDistanceY.get());
+    }
+    
+    public XYPair getRobotOrientedTotalDistanceTraveled() {
+        // if we are facing 90 degrees, no change.
+        // if we are facing 0 degrees (right), this rotates left by 90. Makes sense - if you rotate right, you want
+        // your perception of distance traveled to be that you have gone "leftward."
+        return getTravelVector().rotate(-(currentHeading.getValue() - 90)).clone();
```
WasabiFan: This math should be simplified. Right now it's difficult to understand.

src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+    private ContiguousHeading lastImuHeading;
+    
+    // These are two common robot starting positions - kept here as convenient shorthand.
+    public static final double FACING_AWAY_FROM_DRIVERS = 90;
+    public static final double FACING_TOWARDS_DRIVERS = -90;
+    
+    private final DoubleProperty currentPitch;
+    private final DoubleProperty currentRoll;
+    
+    private final DoubleProperty inherentRioPitch;
```
WasabiFan: Is this actually about the RIO? Or is it the IMU?

src/xbot/common/subsystems/PoseSubsystem.java
```diff
@@ -0,0 +1,184 @@
...
+    private final DoubleProperty inherentRioRoll;
+    
+    private double previousLeftDistance;
+    private double previousRightDistance;
+    
+    private BooleanProperty rioRotated;
```
WasabiFan: This property name doesn't explain what it is. Of course it's rotated -- that's what all the other properties are for. This property seems to define whether pitch and roll are swapped.

src/xbot/common/subsystems/PoseSubsystem.java
aschokking: In this instance at the subsystem level this feels like a reasonable way to me to ensure this history variable is populated before anything that uses it goes. Otherwise we'd need something like a nullable double? or something to test against before reading it and then having the initialization logic there which feels net messier to me.

PR #50, opened 2017-02-05
-If isOnTarget is called before calculate, it will always return FALSE

JohnGilb:
-created override for isOnTarget that also accepts a derivative term

### DIFF
--- a/src/xbot/common/math/PID.java
+++ b/src/xbot/common/math/PID.java
@@ -15,12 +15,9 @@ public class PID
     private double m_targetInputValue;
     private double m_currentInputValue;
     
-    private boolean calculateHasBeenCalled = false;
     private double m_derivativeValue;
-    
     private boolean errorIsSmall = false;
     private boolean derivativeOfErrorIsSmall = false;
-    
     double errorTolerance = -1;
     double errorDerivativeTolerance = -1;
 
@@ -83,9 +80,7 @@ public class PID
      */
     public double calculate(double goal, double current,
             double p, double i, double d, double f)
-    {
-        calculateHasBeenCalled = true;
-        
+    {        
         m_targetInputValue = goal;
         m_currentInputValue = current;
         double result;
@@ -147,22 +142,24 @@ public class PID
             double p, double i, double d) {
         return calculate(goal, current, p, i, d, 0);
     }
-
-    /**
-     * This tells you if the controller is near its goal. 
-     * 
-     * You need to call setErrorTolerance() for this to work well!
-     */
-    public boolean isErrorBelowTolerance() {
-        return errorIsSmall;
-    }
     
     /**
-     * This tells you if the error in the controller is changing very slowly
-     * 
-     * You need to call setErrorDerivativeTolerance() for this to work well!
+     * Check for all conditions where the tolerance is above 0. Since tolerances default to
+     * -1, it will skip any unassigned ones.
      */
-    public boolean isDerivativeOfErrorBelowTolerance() {
-        return derivativeOfErrorIsSmall;
+    public boolean isOnTarget() {
+        boolean checkError = errorTolerance > 0;
+        boolean checkDerivative = errorDerivativeTolerance > 0;
+        
+        boolean isOnTarget = true;
+        
+        if (checkError) {
+            isOnTarget &= errorIsSmall;
+        }
+        if (checkDerivative) {
+            isOnTarget &= derivativeOfErrorIsSmall;
+        }
+        
+        return isOnTarget;
     }
 }
--- a/src/xbot/common/math/PID.java
+++ b/src/xbot/common/math/PID.java
@@ -17,9 +17,9 @@ public class PID
     
     private double m_derivativeValue;
     private boolean errorIsSmall = false;
-    private boolean derivativeOfErrorIsSmall = false;
+    private boolean derivativeIsSmall = false;
     double errorTolerance = -1;
-    double errorDerivativeTolerance = -1;
+    double derivativeTolerance = -1;
 
     /**
      * Resets the PID controller.
@@ -29,7 +29,7 @@ public class PID
         m_prevError = 0;
         m_totalError = 0;
         errorIsSmall = false;
-        derivativeOfErrorIsSmall = false;
+        derivativeIsSmall = false;
     }
     
     /**
@@ -39,7 +39,7 @@ public class PID
      * This is in the same units as your current and goal values.
      */
     public void setErrorTolerance(double errorTolerance) {
-        setTolerances(errorTolerance, this.errorDerivativeTolerance);
+        setTolerances(errorTolerance, this.derivativeTolerance);
     }
     
     /**
@@ -56,9 +56,9 @@ public class PID
         setTolerances(this.errorTolerance, errorDerivativeTolerance);
     }
     
-    public void setTolerances(double errorTolerance, double errorDerivativeTolerance) {
+    public void setTolerances(double errorTolerance, double derivativeTolerance) {
         this.errorTolerance = errorTolerance;
-        this.errorDerivativeTolerance = errorDerivativeTolerance;
+        this.derivativeTolerance = derivativeTolerance;
     }
 
     /**
@@ -118,7 +118,7 @@ public class PID
         result = m_result;
         
         errorIsSmall = Math.abs(m_targetInputValue - m_currentInputValue) < errorTolerance;
-        derivativeOfErrorIsSmall = Math.abs(m_derivativeValue) < errorDerivativeTolerance;
+        derivativeIsSmall = Math.abs(m_derivativeValue) < derivativeTolerance;
 
         return result;
     }
@@ -149,7 +149,7 @@ public class PID
      */
     public boolean isOnTarget() {
         boolean checkError = errorTolerance > 0;
-        boolean checkDerivative = errorDerivativeTolerance > 0;
+        boolean checkDerivative = derivativeTolerance > 0;
         
         boolean isOnTarget = true;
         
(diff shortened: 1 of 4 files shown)


src/xbot/common/math/PIDPropertyManager.java
```diff
@@ -18,6 +21,11 @@ public PIDPropertyManager(String functionName, XPropertyManager propMan,
         propI = propMan.createPersistentProperty(functionName + " I", defaultI);
         propD = propMan.createPersistentProperty(functionName + " D", defaultD);
         propF = propMan.createPersistentProperty(functionName + " F", defaultF);
+        
+        propErrorThreshold = 
+                propMan.createPersistentProperty(functionName + " Error threshold", -1);
```
aschokking: should have an overloaded constructor to set defaults on these?

JohnGilb: Good idea.

src/xbot/common/math/PID.java
```diff
@@ -44,7 +80,7 @@ public void reset()
      */
     public double calculate(double goal, double current,
             double p, double i, double d, double f)
-    {
+    {
```
WasabiFan: Remove whitespace

src/xbot/common/math/PID.java
```diff
@@ -79,6 +116,9 @@ public double calculate(double goal, double current,
             m_result = m_minimumOutput;
         }
         result = m_result;
+        
+        errorIsSmall = Math.abs(m_targetInputValue - m_currentInputValue) < errorTolerance;
```
WasabiFan: If you change the tolerance between when this runs and when `isOnTarget` runs, that method will return nonsense.

src/xbot/common/math/PIDManager.java
```diff
@@ -14,17 +14,30 @@
     private DoubleProperty maxOutput;
     private DoubleProperty minOutput;
     private BooleanProperty isEnabled;
-
-    public PIDManager(String functionName, XPropertyManager propMan, double defaultP, double defaultI, double defaultD,
-            double defaultMaxOutput, double defaultMinOutput) {
-        super(functionName, propMan, defaultP, defaultI, defaultD, 0);
+    
+    public PIDManager(String functionName, XPropertyManager propMan, double defaultP, double defaultI, double defaultD, double defaultF,
```
WasabiFan: Separate these onto individual lines so it's legible.

src/xbot/common/math/PID.java
```diff
@@ -102,17 +142,24 @@ public double calculate(double goal, double current,
...
+     * -1, it will skip any unassigned ones.
      */
-    public boolean isOnTarget(double absoluteTolerance)
-    {
-        return Math.abs(m_targetInputValue - m_currentInputValue) < absoluteTolerance;
+    public boolean isOnTarget() {
+        boolean checkError = errorTolerance > 0;
+        boolean checkDerivative = derivativeTolerance > 0;
+        
+        boolean isOnTarget = true;
```
WasabiFan: If there isn't a tolerance set, it defaults to true? That's not what I would expect. If I implemented a command which would use this as the value for `isFinished`, it would immediately exit if I didn't have a tolerance set.

src/xbot/common/math/PID.java
```diff
@@ -23,6 +28,37 @@ public void reset()
...
+        derivativeIsSmall = false;
+    }
+    
+    /**
+     * Set how close the error can be before it is considered
+     * "on-target."
+     *            
+     * This is in the same units as your current and goal values.
+     */
+    public void setErrorTolerance(double errorTolerance) {
```
WasabiFan: The fact that you can set the value to a negative number and have it silently ignore this value is worrying. Can't we use a `Double` instead and set it to `null`?

src/xbot/common/math/PID.java
```diff
@@ -23,6 +28,37 @@ public void reset()
...
+        setTolerances(errorTolerance, this.derivativeTolerance);
```
WasabiFan: This should just set the property manually. You aren't gaining anything by calling the other method, and this adds an extra layer of indirection for a reader. Additionally, if we later wanted to add logic to `setTolerances` that reset things when the value was changed, this would make that logic needlessly complicated.

src/xbot/common/math/PID.java
JohnGilb: @WasabiFan  Since the error tolerance are linked to DoubleProperties, we're constrained by what we can interact with on the SmartDashboard.

I wanted the ability to turn constraint checking on and off via the SmartDashboard, and to control the constraints themselves. Combining them so that negative values cause constraint checking to be skipped helps reduce clutter a bit.

On the other hand, we currently have two constraints, and may add a third (time). Adding three more enable/disable properties would seem kind of messy to display on the SD. Is there a better solution?

src/xbot/common/math/PID.java
JohnGilb: The PID class currently is just a bunch of math - it has no visibility into properties.

src/xbot/common/math/PID.java
JohnGilb: configured my editor, now I can actually see this =]

src/xbot/common/math/PID.java
JohnGilb: If you're calling isOnTarget and you didn't set a tolerance, we can do two simple things:
-Always say true. Command ends immediately.
-Always say false. Command never ends.

Both of these are bad - but by having the command immediately end, we actually highlight the problem faster. If you were doing tests of a command on the robot, we could notice quickly that the command was starting but stopping immediately, which would let us to investigate isFinished.

There's probably some more advanced stuff we could do to try and highlight this to the programmer (maybe a test-only exception, and/or error logging), but for what it does on the robot I think we have to choose one of the two approaches above.

Do you have any feedback on which approach is more desirable?

src/xbot/common/math/PIDManager.java
JohnGilb: good idea.

src/xbot/common/math/PID.java
WasabiFan: Language properties -- fields -- not dashboard properties.

```java
this.errorTolerance = errorTolerance;
```

src/xbot/common/math/PID.java
WasabiFan: I can imagine a functional case where we'd want a command using this logic to run indefinitely: you have a button that rotates the robot to a certain angle and stays there, On the other hand, having it end _immediately_ doesn't have a use case -- it doesn't do anything. To me, it seems like it's a lot easier to figure out the answer to "why is my robot stuck trying to point in this direction" than "why did my robot not point where it was supposed to?". If it never stops, at least you can see it live on the dashboard; if it exits immediately, you have to find the info in logs and realize that something finished too early.

src/xbot/common/math/PID.java
WasabiFan: I think we should make our programmatic interfaces as natural as possible, and then take the hit in the dashboard interaction. If we have to special-case a value (whether it's negative numbers, 0, or something else) I'd rather do that in the command or subsystem when taking data from the related property; that way, we can at least minimize bugs when using these APIs in code.

src/xbot/common/math/PID.java
aschokking: In the Command control loop `isFinished()` is called before `execute()`, I'm not sure there's much room for a race condition here and it would just be while tuning the tolerance in any event?

src/xbot/common/math/PID.java
aschokking: In general I don't think we'll be using these as a way to configure if a command should even finish or not. That said, I agree it's probably reasonable to say if no tolerances are set that it never finishes. 

We could preface this section with a test for any of the tolerances being there and if none return `false`?

src/xbot/common/math/PIDManager.java
aschokking: I thought we had checkstyle looking out for really long lines but maybe not.

JohnGilb: It's hard to tell from the github view, but it was already broken up into a few line segments, so checkstyle was not being hit. It's since been put into individual lines per argument

src/xbot/common/math/PID.java
JohnGilb: Sounds like the two of you agree that False is the better error case - I'm willing to go with that.

src/xbot/common/math/PID.java
JohnGilb: Sounds reasonable. I'll split it into enable and values and try it on the robot for a while, see how the flow feels.

(In this particular case I think it makes sense to consider SD a major consumer, since almost all of our PID values are tuned/modified using it.)

I'm not sure what the proper response is to negative values being set programmatically or via the SD. Do we want to intercept and invert them? Silently ignore, falling into the "never finishes" condition?

JohnGilb: For tests, we can use the RobotAssertionManager to make sure nobody is setting negative values by throwing exceptions in their tests, but not sure what the best runtime option is. Might just be silently continue (akin to when you isOnTarget with no tolerances?)

src/xbot/common/math/PID.java
WasabiFan: Regardless of whether the current implementation of one caller would _probably_ avoid it, it's faulty logic and fundamentally a bug. I could easily imagine this case coming up in tests or while tuning parameters, and a robot doing unpredictable things is never a good thing. The fix is simple and should be made.

JohnGilb: What's the fix?

WasabiFan: If the tolerance checks aren't enabled, set `errorIsSmall` (and its derivative counterpart) to be `false`.

PR #51, opened 2017-02-05
Improvements to PIDManager re: Feedback

JohnGilb:
-Added assertion check (for tests) that we're not using negative tolerances
-Split enable/disable and tolerance values into separate properties
-Added tests for these behaviors

### DIFF
--- a/src/xbot/common/math/PID.java
+++ b/src/xbot/common/math/PID.java
@@ -124,8 +124,15 @@ public class PID
         }
         result = m_result;
         
-        errorIsSmall = Math.abs(m_targetInputValue - m_currentInputValue) < errorTolerance;
-        derivativeIsSmall = Math.abs(m_derivativeValue) < derivativeTolerance;
+        errorIsSmall = false;
+        if (checkErrorThreshold) {
+            errorIsSmall = Math.abs(m_targetInputValue - m_currentInputValue) < errorTolerance;
+        }
+        
+        derivativeIsSmall = false;
+        if (checkDerivativeThreshold) {
+            derivativeIsSmall = Math.abs(m_derivativeValue) < derivativeTolerance;
+        }
 
         return result;
     }
--- a/src/xbot/common/math/PID.java
+++ b/src/xbot/common/math/PID.java
@@ -18,9 +18,12 @@ public class PID
     private double m_derivativeValue;
     private boolean errorIsSmall = false;
     private boolean derivativeIsSmall = false;
-    double errorTolerance = -1;
-    double derivativeTolerance = -1;
-
+    double errorTolerance = 0;
+    double derivativeTolerance = 0;
+    
+    private boolean checkErrorThreshold = false;
+    private boolean checkDerivativeThreshold = false;
+    
     /**
      * Resets the PID controller.
      */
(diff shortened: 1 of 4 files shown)


src/xbot/common/math/PID.java
```diff
@@ -18,9 +18,12 @@
     private double m_derivativeValue;
     private boolean errorIsSmall = false;
     private boolean derivativeIsSmall = false;
-    double errorTolerance = -1;
-    double derivativeTolerance = -1;
-
+    double errorTolerance = 0;
```
WasabiFan: What's your thinking with using a separate Boolean instead of `null`?

src/xbot/common/math/PIDPropertyManager.java
```diff
@@ -14,7 +17,12 @@
     private final DoubleProperty propErrorThreshold;
     private final DoubleProperty propDerivativeThreshold;
     
-    public PIDPropertyManager(String functionName, XPropertyManager propMan, 
+    private final BooleanProperty propEnableErrorThreshold;
+    private final BooleanProperty propEnableDerivativeThreshold;
+    
+    private final RobotAssertionManager assertionManager;
+    
+    public PIDPropertyManager(String functionName, XPropertyManager propMan, RobotAssertionManager assertionManager,
```
WasabiFan: Wrap this constructor

src/xbot/common/math/PIDPropertyManager.java
```diff
@@ -70,14 +85,32 @@ public double getErrorThreshold() {
...
     
     public double getDerivativeThreshold() {
         return propDerivativeThreshold.get();
     }
     
-    public void setDerivativeThreshold(double errorThreshold) {
-        propDerivativeThreshold.set(errorThreshold);
-    }    
+    public void setDerivativeThreshold(double derivativeThreshold) {
+        assertionManager.assertTrue(derivativeThreshold >= 0, "Thresholds won't work if they are negative!");
```
WasabiFan: If it's a negative number and we are running on-robot, I think we should take the absolute value so it's at least sane.

src/xbot/common/math/PID.java
aschokking: You mean using a `Double` instead so it would be nullable?

WasabiFan: Yes; you need a reference type to make it `null`.

JohnGilb: I want to control enable/disable via the SmartDashboard. Since that's already a boolean, I just pass that information down into the PID class.

src/xbot/common/math/PIDPropertyManager.java
JohnGilb: What does that mean?

src/xbot/common/math/PIDPropertyManager.java
JohnGilb: Sounds good.

src/xbot/common/math/PIDPropertyManager.java
WasabiFan: Sorry, as in "line wrapping": put each parameter on its own line.

src/xbot/common/math/PID.java
```diff
@@ -113,8 +124,15 @@ public double calculate(double goal, double current,
         }
         result = m_result;
         
-        errorIsSmall = Math.abs(m_targetInputValue - m_currentInputValue) < errorTolerance;
-        derivativeIsSmall = Math.abs(m_derivativeValue) < derivativeTolerance;
+        errorIsSmall = false;
```
WasabiFan: This logic can be simplified to:

```java
errorIsSmall = checkErrorThreshold && Math.abs(m_targetInputValue - m_currentInputValue) < errorTolerance
```

JohnGilb: concise!

PR #52, opened 2017-02-05
Make PoseSubsystem into AbstractPoseSubsystem

JohnGilb:
Given that each year's robot is going to have a different way of computing leftDistance and rightDistance, the PoseSubsystem is made abstract, and each year's implementation will have a solution to getting that information.

### DIFF
--- a/src/xbot/common/subsystems/AbstractPoseSubsystem.java
+++ b/src/xbot/common/subsystems/AbstractPoseSubsystem.java
@@ -5,6 +5,7 @@ import org.apache.log4j.Logger;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 
+import javafx.util.Pair;
 import xbot.common.command.BaseSubsystem;
 import xbot.common.controls.sensors.NavImu.ImuType;
 import xbot.common.controls.sensors.XGyro;
@@ -15,10 +16,9 @@ import xbot.common.properties.BooleanProperty;
 import xbot.common.properties.DoubleProperty;
 import xbot.common.properties.XPropertyManager;
 
-@Singleton
-public class PoseSubsystem extends BaseSubsystem {
+public abstract class AbstractPoseSubsystem extends BaseSubsystem {
         
-    private static Logger log = Logger.getLogger(PoseSubsystem.class);
+    private static Logger log = Logger.getLogger(AbstractPoseSubsystem.class);
     public final XGyro imu;
     
     private final DoubleProperty leftDriveDistance;
@@ -47,8 +47,7 @@ public class PoseSubsystem extends BaseSubsystem {
     
     private BooleanProperty rioRotated;
     
-    @Inject
-    public PoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
+    public AbstractPoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
         log.info("Creating PoseSubsystem");
         imu = factory.getGyro(ImuType.navX);
         
@@ -152,8 +151,11 @@ public class PoseSubsystem extends BaseSubsystem {
      */
     public void updatePose(double leftDistance, double rightDistance) {
         updateCurrentHeading();
-        updateDistanceTraveled(leftDistance, rightDistance);
+        updateDistanceTraveled(getLeftDriveDistance(), getRightDriveDistance());
     }
+    
+    protected abstract double getLeftDriveDistance();
+    protected abstract double getRightDriveDistance();
     
     public double getRobotPitch() {
         return getRealPitch() - inherentRioPitch.get();
--- a/src/xbot/common/subsystems/AbstractPoseSubsystem.java
+++ b/src/xbot/common/subsystems/AbstractPoseSubsystem.java
@@ -149,7 +149,7 @@ public abstract class AbstractPoseSubsystem extends BaseSubsystem {
      * @param leftDistance total distance, in inches, that the left drivetrain has traveled
      * @param rightDistance total distance, in inches, that the right drivetrain has traveled
      */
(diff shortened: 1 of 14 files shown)


tests/xbot/common/subsystems/TestPoseSubsystem.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Override
+    protected double getLeftDriveDistance() {
+        return left.getPosition();
+    }
+
+    @Override
+    protected double getRightDriveDistance() {
+        // TODO Auto-generated method stub
```
WasabiFan: Remove TODO

JohnGilb: good catch, thanks

src/xbot/common/subsystems/AbstractPoseSubsystem.java
```diff
@@ -150,11 +149,14 @@ public void setCurrentHeading(double headingInDegrees){
      * @param leftDistance total distance, in inches, that the left drivetrain has traveled
```
aschokking: should remove params from javadoc comment

tests/xbot/common/subsystems/MockPoseSubsystem.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.google.inject.Inject;
+
+import xbot.common.controls.actuators.XCANTalon;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.properties.XPropertyManager;
+
+public class MockPoseSubsystem extends AbstractPoseSubsystem {
+
+    private XCANTalon left;
```
aschokking: Ignore-able suggestion, but since this Mock class is just about getting/setting the left/right distance, maybe just have methods that do that instead of needing to involve XCANTalon at all?

tests/xbot/common/subsystems/MockPoseSubsystem.java
```diff
@@ -0,0 +1,34 @@
+package xbot.common.subsystems;
+
+import com.google.inject.Inject;
+
+import xbot.common.controls.actuators.XCANTalon;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.properties.XPropertyManager;
+
+public class MockPoseSubsystem extends AbstractPoseSubsystem {
```
WasabiFan: I'd rather this be called `TestPoseSubsystem`, or something other than "mock". We use the term "mock" to refer to fake entities that aren't legitimately functional.

src/xbot/common/subsystems/AbstractPoseSubsystem.java
```diff
@@ -47,8 +47,7 @@
     
     private BooleanProperty rioRotated;
     
-    @Inject
-    public PoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
+    public AbstractPoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
```
aschokking: I'm not wildly in love with putting Abstract in the class name, I think we've tended towards `Base` in these situations so far in the commonlib?

src/xbot/common/subsystems/AbstractPoseSubsystem.java
JohnGilb: Good point!

src/xbot/common/subsystems/AbstractPoseSubsystem.java
JohnGilb: good idea - consistent.

src/xbot/common/subsystems/BasePoseSubsystem.java
```diff
@@ -47,8 +47,7 @@
     
     private BooleanProperty rioRotated;
     
-    @Inject
-    public PoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
+    public BasePoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
```
aschokking: Does it make sense for this to be a TelemetrySource now since the drive part can now be a poll?

JohnGilb: Good point!

WasabiFan: If this class doesn't actually publish any telemetry data (it doesn't right now) it shouldn't be a `TelemetrySource`. A subclass can decide to implement that interface, but there's no reason to require it here.

JohnGilb: Ah - I neglected to actually override the method and call updatePose(). Added that now.

WasabiFan: I'm still hesitant to use `TelemetrySource` like this, for two reasons:

- Implementing this interface doesn't actually cause that method to be called; the season-specific code must manually register it.
- The telemetry infrastructure is intended to be for things that are for human use when debugging, but aren't critical to the function of the system; or at least, that is what I had expected it to be. By putting the pose update logic here, we're putting a lot of important functionality in the hands of that debugging-related system.

aschokking: I think `TelemetrySource` just might be a bad name for what's going on here, there's a class of things we want to run every robot tick regardless of whats going on in terms of enabled/disabled (one of which is updating telemetry on the dashboard but another is this kind of sensor integration). Open to better names for `TelemetrySource` given the broader set of scenarios it's really meant for.

JohnGilb: `PeriodicDataSource`? =]

JohnGilb:
Finally, the tests can pass =]

PR #54, opened 2017-02-06
Guice-up PIDManager creation

aschokking:
Make it easier to create a PIDManager with a factory implemented using guice's AssistedInjection.

Updated the PIDManagerTest with how to use it.

This has the advantage of not requiring consumers to provide all of the injectable dependencies PIDManager has now or in the future. Also general good practice in how AssistedInjection works to help us see if there are other places we might want to leverage it.

### DIFF
--- a/src/xbot/common/math/PIDManagerFactory.java
+++ b/src/xbot/common/math/PIDManagerFactory.java
@@ -1,17 +1,45 @@
 package xbot.common.math;
 
+import com.google.inject.assistedinject.Assisted;
+
 public interface PIDManagerFactory {
 
-    public PIDManager create(String functionName, double defaultP, double defaultI, double defaultD, double defaultF,
-            double defaultMaxOutput, double defaultMinOutput, double errorThreshold, double derivativeThreshold);
+    public PIDManager create(
+            String functionName,
+            @Assisted("defaultP") double defaultP, 
+            @Assisted("defaultI") double defaultI, 
+            @Assisted("defaultD") double defaultD, 
+            @Assisted("defaultF") double defaultF,
+            @Assisted("defaultMaxOutput") double defaultMaxOutput, 
+            @Assisted("defaultMinOutput") double defaultMinOutput,
+            @Assisted("errorThreshold") double errorThreshold, 
+            @Assisted("derivativeThreshold") double derivativeThreshold);
+    
+    public PIDManager create(
+            String functionName,
+            @Assisted("defaultP") double defaultP, 
+            @Assisted("defaultI") double defaultI, 
+            @Assisted("defaultD") double defaultD, 
+            @Assisted("defaultF") double defaultF,
+            @Assisted("defaultMaxOutput") double defaultMaxOutput, 
+            @Assisted("defaultMinOutput") double defaultMinOutput);
 
-    public PIDManager create(String functionName, double defaultP, double defaultI, double defaultD, double defaultF,
-            double defaultMaxOutput, double defaultMinOutput);
+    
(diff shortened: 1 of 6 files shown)


src/xbot/common/math/PIDManagerFactory.java
```diff
@@ -0,0 +1,45 @@
+package xbot.common.math;
+
+import com.google.inject.assistedinject.Assisted;
+
+public interface PIDManagerFactory {
```
JohnGilb: I can see how writing the factory here plus the constructors elsewhere is kind of a pain, but man, what a time savings elsewhere.

WasabiFan:
Although this does make creation of these objects more concise in some cases, it feels wrong to make creation of this particular object different from the rest of the utilities that we have. It's also odd to make it so that proper creation of this object requires "special knowledge," likely leading to fragmentation in any consuming codebases.

Looking specifically at code savings, I see very little benefit overall. In doing this, you add a `injector.getInstance` call to save yourself two of them; a net gain of calling that method one less time. Given that you must define a variable to store the factory object, you aren't making it any more concise, and you're diverting the focus away from the logic by adding extra injection objects that aren't involved in the task at hand. That's in the worst case. In many situations, you already have a property manager, so you are making your code _longer_ by using this technique.

It seems like there isn't a reason for us to do this, outside of testing the `PIDManager` class itself. I don't see enough to justify merging these changes.

JohnGilb:
The real advantage comes from when we want to add some new injectable to the PIDManager. When I added the RobotAssertionManager, I had to go through and change a ton of files to add that to the constructor of every single class that was using the PIDManager.

With this approach, any new "injectable" we add to PIDManager is invisible to all its consumers. That's really nice from a refactoring standpoint! =]

WasabiFan:
That's a good point. What is it that prevents us from having the class itself use the injector? Why can't we have it create these classes? Or, what if we created a class like `PIDManager.Dependencies` that included the `RobotAssertionManager`, `XPropertyManager`, etc. and injected that instead (so we just passed the single injected object)? Each of these options would give us the benefit of the assisted injection while keeping the logic localized -- my main goal is to remove the requirement for the calls in the injection modules.

src/xbot/common/math/PIDManagerFactory.java
aschokking: Agreed, and mostly just because PIDManager has a bajillion different constructors which we don't often do. That said for this kind of thing feels worth the initial investment of time on the CommonLib side to ease things at competition bot time for sure.

aschokking:
It's a good question. The rub is wanting to have a constructor that takes non-injectable parameters. 

For such cases there isn't really a way around having a factory (like we manually did with WPIFactory). If PIDManager still wants to have such parameters in it's constructor than it can't be injected normally AND it won't be able to use method injection either sadly. 

You could group all the dependencies together into a class that people who want to create a PIDManager get injected instead but that required knowledge (that you need to ask for PMDep class in injector) doesn't seem a lot more elegant than a factory.

aschokking:
Sorry meant to speak to your other suggestion as well, having PIDManager get the injector is really icky unfortunately. You really want at any given time there to be only 1 injector (so for instance when writing unit tests you know all the Singletons will do what you think) so you need some way of plumbing it around where you can control things but it can't be inject itself. In the past we've had a static version somewhere but then it makes the unit tests tricky. 

Our use of it in Robot.java is already not very elegant and I've been trying to think of ways to flip it around so the Robot itself is injectable so we can just use constructor/method injection there like we do in the rest of the robot.

PR #57, opened 2017-02-10
Add ability to temporarily mask I value

WasabiFan:


aschokking:
What's the use case for this?

JohnGilb:
Is this what you were talking about where you didn't want to consider the k_I value when you temporarily lose vision of the target, but not reset the accumulator?

src/xbot/common/math/PIDManager.java
```diff
@@ -19,6 +19,7 @@
     private DoubleProperty maxOutput;
     private DoubleProperty minOutput;
     private BooleanProperty isEnabled;
+    private boolean isIMasked = false;
```
JohnGilb: I feel like `mask` is slightly ambiguous, I don't get an immediate sense of "masked on" or "masked off". 

What about `isISupressed` or some other verb more closely associated with hiding/suppressing/disabling?

JohnGilb: It's funny because in normal English people would absolutely get the idea that "masking" is "hiding", but in CS it's a little less clear =]

aschokking: I agree this is kind of tricky to grok on face value, I like suppressed a bit more but don't feel strongly about it.

aschokking:
On a more macro level is this something we'd expect to use in a lot of PIDs and if not maybe we could subclass it in this years code?

WasabiFan:
@JohnGilb got the intent correct. The idea is to add a logical way of "pausing" for a bit if tracking info is lost or some other precondition isn't met. There may be a better approach to this, maybe one that decays the `I` value as well. Open to suggestions.

aschokking:
Have you bounced this problem past Jeff? It sounded like he had done a lot in this space before.

WasabiFan:
I was actually intent on closing this; I modified my code to not need it because I decided it wasn't the proper solution.

PR #58, opened 2017-02-12
Add PIDPropertyManager to the factory

aschokking:


src/xbot/common/math/PIDManagerFactory.java
```diff
@@ -4,7 +4,7 @@
 
 public interface PIDManagerFactory {
```
JohnGilb: I'm kiiiiiind of leaning towards `PIDFactory` if it makes a bunch of different PID-related things? No strong feelings though.

JohnGilb: or maybe `MathFactory`, since they're all in that package? `XbotMathFactory`?

PR #59, opened 2017-02-12
Allow Base Subsystem/Command to have 'name' set explicitly 

aschokking:


lib/log4jConfig/log4j.xml
```diff
@@ -6,15 +6,15 @@
       class="org.apache.log4j.ConsoleAppender">
       <param name="Threshold" value="INFO" />
       <layout class="org.apache.log4j.PatternLayout">
-         <param name="ConversionPattern" value="%5p | %d | %-45.45F | %L | %m%n" />
+         <param name="ConversionPattern" value="%5p | %d | %L | %-25c{1} | %m%n" />
```
JohnGilb: Pad to 40! =]

PR #60, opened 2017-02-12
Changed the buttons as they were not properly configured

Unagifan:
Sticks on Gamepad were not correctly configured.

### DIFF
--- a/src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
+++ b/src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
@@ -53,7 +53,7 @@ public class XboxControllerWpiAdapter implements XXboxController {
     
     public double getRightStickX(){
         
-        return internalXboxController.getX(Hand.kRight);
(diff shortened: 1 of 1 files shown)


src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
```diff
@@ -53,7 +53,7 @@ public double getLeftStickX(){
     
     public double getRightStickX(){
         
-        return internalXboxController.getX(Hand.kRight);
+        return internalXboxController.getRawAxis(3);
```
WasabiFan: Either all stick getters should use the `getX`/`getY` calls or all should use the raw axis version. Mixing them makes this logic significantly harder to follow.

aschokking: Sounds reasonable, created a follow up task: https://app.asana.com/0/38541457243752/270403025670257

PR #61, opened 2017-02-12
Feature/create base srx controlled subsystem

JohnGilb:


src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
```diff
@@ -0,0 +1,180 @@
...
+    
+    public boolean isAtSpeed() {
+        return Math.abs(getSpeed() - systemTargetSpeed.get()) <= systemSpeedThresh.get();
+    }
+    
+    public double getTargetSpeed() {
+        return systemTargetSpeed.get();
+    }
+
+    @Override
```
aschokking: do you still do `@Override` for interface methods that aren't shadowing an existing method?

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
```diff
@@ -0,0 +1,180 @@
...
+     */
+    public BaseSRXControlledSubsystem(
+            String systemName,
+            int masterChannel,
+            boolean invertMaster,
+            boolean invertMasterSensor,
+            WPIFactory factory, 
+            PIDPropertyManager pidPropertyManager,
+            XPropertyManager propManager){
+        log.info("Creating " + systemName + " system");
```
aschokking: Call `super(name);`

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
```diff
@@ -0,0 +1,180 @@
...
+        
+        this.pidPropertyManager = pidPropertyManager;
+        this.systemName = systemName;
```
aschokking: reason for a new name property?

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
```diff
@@ -0,0 +1,180 @@
...
+        masterMotor.setInverted(motorInverted);
+        masterMotor.reverseSensor(motorSensorInverted);
+        masterMotor.setControlMode(TalonControlMode.Speed);
+        masterMotor.setProfile(0);
+        
+        setupExtendedMotorProperties();
+    }
+        
+    private void updateMotorPidValues() {
+        masterMotor.setP(pidPropertyManager.getP());
```
aschokking: maybe a comment about when this should be called? Should it happen during periodic?

aschokking: ah nvm, just noticed where it's called and that it's private, alls well. Maybe would be fine to be protected though in case someone needs to tweak it?

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
```diff
@@ -0,0 +1,218 @@
...
+    protected final DoubleProperty systemCurrentSpeed;
+    protected final DoubleProperty systemTargetSpeed;
+    protected final DoubleProperty systemOutputPower;
+    protected final DoubleProperty systemTalonError;
+    protected final BooleanProperty atSpeedProp;
+    protected final BooleanProperty enablesystemLogging;
+    
+    private final DoubleProperty systemSpeedThresh;
+    protected final PIDPropertyManager pidPropertyManager;
+    protected final String systemName;
```
aschokking: delete me!

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
```diff
@@ -0,0 +1,205 @@
...
+     * 
+     * @param systemName What the system is called. This will apply to various Properties.
+     * @param masterChannel The CAN index of the master motor (or the only motor, for a simple system)
+     * @param followChannel The CAN index of the follow motor (-1 if no follow motor)
+     * @param factory The WPIFactory
+     * @param pidPropertyManager The default PIDF values the system should use
+     * @param propManager The XPropertyManager
+     */
+    public BaseSRXControlledSubsystem(
+            String systemName,
```
aschokking: let's rename this just `name` to be consistent with Subsystem?

WasabiFan:
We've recently been adding a lot of "base subsystems" for various tasks. While I think that is good in small amounts, taking it to this extreme is starting to get worrying. It might make implementing the most basic systems easier, but it makes the interfaces less natural (because they are forced to be generic) and makes things intensely difficult if you need to modify the default behavior.

In this case, we have a subsystem base class that claims to be a base for any Talon SRX-controlled subsystem, yet it is really a _speed maintainer subsystem_ that operates on a pair of CAN Talons. Calling it the more accurate `BaseCANTalonPairSpeedMaintainerSubsystem` begins to get so specific that I don't see it being a net gain -- we rarely need more than one such system in a year, if at all. I'd rather be able to copy this logic between years and modify it to fit the specific scenario than have to work around this monolithic base for the one time it's needed.

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
JohnGilb: Time to find out what happens if you don't.

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
JohnGilb: yes, can just use the base class name now

src/xbot/common/subsystems/BaseSRXControlledSubsystem.java
JohnGilb: sure!

JohnGilb:
This is mostly motivated by the fact our Belt and Shooter systems are just copy-pastes of each other already.

I agree with the rename.

I think I'm going to move this logic to this year's competition code for now - if we find ourselves wanting to take this approach next year as well, then we can consider moving it to the common codebase.

PR #64, opened 2017-02-17
Feature/add xxbox joystick inversion

SethNguyenXBOT:


src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -108,5 +175,51 @@ public XboxController getInternalController() {
...
+    @Override
+    public void setLeftStickXInversion(boolean inverted) {
+        // TODO Auto-generated method stub
+        xLeftInverted = inverted;
+    }
 
-}
+    @Override
+    public void setLeftStickYInversion(boolean inverted) {
+        // TODO Auto-generated method stub
```
WasabiFan: Remove TODOs

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -108,5 +175,51 @@ public XboxController getInternalController() {
...
+        yLeftInverted = inverted;
+    }
+    
+    @Override
+    public boolean getRightStickXInversion() {
+        // TODO Auto-generated method stub
```
WasabiFan: Implement these

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -12,10 +15,74 @@ public MockXboxController(int port) {
...
     boolean[] buttonValues = new boolean[10];
     
+    public void setRightX(double x) {
+        double value = x * (xRightInverted ? -1 : 1);
+        this.x = value;
+        setRawAxis(0, value);
+    }
+
+    public void setLeftX(double x) {
+        double value = x * (xLeftInverted ? -1 : 1);
```
WasabiFan: What's the thinking behind inverting values that are passed in?

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -12,10 +15,74 @@ public MockXboxController(int port) {
...
     double rightTriggerAxis;
+    private boolean xRightInverted = false;
+    private boolean yRightInverted = false;
+    private boolean xLeftInverted = false;
+    private boolean yLeftInverted = false;
+    
+    Map<Integer, Boolean> buttons = new HashMap<Integer, Boolean>();
+    Map<Integer, Double> rawAxis = new HashMap<Integer, Double>();
     
     boolean[] buttonValues = new boolean[10];
```
WasabiFan: Remove this now that you're using a `HashMap`

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -12,10 +15,74 @@ public MockXboxController(int port) {
     XYPair leftJoystickAxis;
     XYPair rightJoystickAxis;
     double leftTriggerAxis;
+    double x = 0;
```
WasabiFan: This shouldn't be needed; you have vectors for both sticks.

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -12,10 +15,74 @@ public MockXboxController(int port) {
...
+
+    public void pressButton(int button) {
+        buttons.put(button, true);
+    }
+
+    public void releaseButton(int button) {
+        buttons.put(button, false);
+    }
+
+    public XYPair getRightVector() {
```
WasabiFan: You have getters that handle the inversion... use them here.

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -12,10 +15,74 @@ public MockXboxController(int port) {
...
+        return new XYPair(this.x * (xRightInverted ? -1 : 1), this.y
+                * (yRightInverted ? -1 : 1));
+    }
+
+    public XYPair getLeftVector() {
+        return new XYPair(this.x * (xLeftInverted ? -1 : 1), this.y
```
WasabiFan: These should use the internal vectors instead of these variables. As-is, both sticks return the same values.

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -12,10 +15,74 @@ public MockXboxController(int port) {
...
+                * (yLeftInverted ? -1 : 1));
+    }
+    
+    public boolean getRawButton(int button) {
+        return getButton(button);
+    }
+    
+    public boolean getButton(int button)
+    {
+        return buttons.getOrDefault(button, false);
```
WasabiFan: You never set this map; all the setters use the duped array.

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -108,5 +175,51 @@ public XboxController getInternalController() {
         return null;
     }
 
+    @Override
+    public void setRightStickXInversion(boolean inverted) {
+        // TODO Auto-generated method stub
```
WasabiFan: Remove TODOs

src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
```diff
@@ -1,12 +1,18 @@
...
 
+import edu.wpi.first.wpilibj.GenericHID;
 import edu.wpi.first.wpilibj.GenericHID.Hand;
 import edu.wpi.first.wpilibj.XboxController;
 import xbot.common.math.XYPair;
 
 public class XboxControllerWpiAdapter implements XXboxController {
     
     final XboxController internalXboxController;
+    private GenericHID internalHID;
```
WasabiFan: Either use the Xbox controller or use this, not both. You should be using the former.

src/xbot/common/controls/sensors/MockXboxController.java
SethNguyenXBOT: What do you mean?

WasabiFan: Let's say I have an instance of this mock class. If I keep inversion off, the value that I `set()` for the virtual value will be equal to the one I `get()`. If I then turn inversion on, _that will still be the case_. That seems potentially confusing to me, given that the general style of the mock class is that the imaginary value which you set is equivalent to the value that would be read from the hardware. I was wondering what the thought process was here.

src/xbot/common/controls/sensors/MockXboxController.java
SethNguyenXBOT: What do you mean?

WasabiFan: There's an array called `buttonValues` and a map called `buttons`. The map is used in some places and the array in others. Instead, all of them should use the map.

src/xbot/common/controls/sensors/MockXboxController.java
SethNguyenXBOT: Changing the setters instead of the getters is much easier at this point so I did them instead.

WasabiFan:
It's worth noting that the current implementation stores the data in individual objects, so creating two controllers on the same port is treated as separate devices; the port parameter is ignored. That means there's no warning for duped controllers.

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -13,15 +16,62 @@ public MockXboxController(int port) {
...
+    public boolean getRawButton(int button) {
+        return getButton(button);
+    }
+    
+    public boolean getButton(int button)
+    {
+        return buttons.getOrDefault(button, false);
+    }
     
     public void setButton(XboxButtons buttonName, boolean pressed) {
```
WasabiFan: If the `getButton` interface has one that takes an int and one that takes a button enum, the `setButton` interface should too. Then pick one of those two versions and call it everywhere else you need to set a button.

src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
```diff
@@ -7,6 +7,10 @@
 public class XboxControllerWpiAdapter implements XXboxController {
     
     final XboxController internalXboxController;
+    private boolean xRightInverted = false;
+    private boolean yRightInverted = false;
+    private boolean xLeftInverted = false;
+    private boolean yLeftInverted = false;
         
     public enum XboxButtons {
```
WasabiFan: Make this singular

src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
```diff
@@ -65,7 +69,46 @@ public double getRightStickY(){
        
        return internalXboxController.getRawAxis(4);
    }
+
+   public boolean getLeftStickXInversion()
```
aschokking: I might just be missing it, but shouldn't these be used in all of the get calls (like `getLeftStickY()`) to actually apply the inversion?

src/xbot/common/controls/sensors/MockXboxController.java
```diff
@@ -90,42 +94,24 @@ public void setRightTriggerPressed(double x){
...
     @Override
     public double getLeftStickY() {
-        return leftJoystickAxis.y;
+        return getLeftVector().y;
     }
 
     @Override
     public double getRightStickY() {
-        return rightJoystickAxis.y;
+        return getRightVector().y;
```
WasabiFan: Invert this, so that the vector getter calls the  individual ones. Every call to `getRightVector` creates a vector object, yet here it's immediately discarded.

src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
JohnGilb: Updated.

src/xbot/common/controls/sensors/MockXboxController.java
JohnGilb: Seems okay as is.

WasabiFan: "seems" is the key word. This code does a heap allocation and then immediately throws away the reference _every time you ask for the stick axis_. That is unquestionably an inversion of the call chain.

src/xbot/common/controls/sensors/XboxControllerWpiAdapter.java
```diff
@@ -36,36 +42,81 @@ public XboxControllerWpiAdapter(int port) {
...
+    
+    public double getLeftStickX(){
        return internalXboxController.getX(Hand.kLeft);
-   }
+    }
     
     public double getRightStickX(){
-        
-        return internalXboxController.getRawAxis(3);
+       return internalXboxController.getRawAxis(3);
```
aschokking: need to update these to use getRightVector() now so they get the inversion too!

PR #65, opened 2017-02-18
Create commands that act upon the pose subsystem. This involves a few…

JohnGilb:
… things:

-Move commands from Competition library to CommonLib
-Have commands depend on BasePoseSubsystem
-Have the injector bind BasePoseSubsystem to TestPoseSubsystem for test

### DIFF
--- a/src/xbot/common/subsystems/pose/BasePoseSubsystem.java
+++ b/src/xbot/common/subsystems/pose/BasePoseSubsystem.java
@@ -1,0 +1,185 @@
+package xbot.common.subsystems.pose;
+
+import org.apache.log4j.Logger;
+
+import xbot.common.command.BaseSubsystem;
+import xbot.common.command.PeriodicDataSource;
+import xbot.common.controls.sensors.NavImu.ImuType;
+import xbot.common.controls.sensors.XGyro;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.math.ContiguousHeading;
+import xbot.common.math.XYPair;
+import xbot.common.properties.BooleanProperty;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+public abstract class BasePoseSubsystem extends BaseSubsystem implements PeriodicDataSource {
+
+    public final XGyro imu;
+    
+    private final DoubleProperty leftDriveDistance;
+    private final DoubleProperty rightDriveDistance;
+    
+    private final DoubleProperty totalDistanceX;
+    private final DoubleProperty totalDistanceY;
+    
+    private ContiguousHeading currentHeading;
+    private final DoubleProperty currentHeadingProp;
+    
+    private ContiguousHeading lastImuHeading;
+    
+    // These are two common robot starting positions - kept here as convenient shorthand.
+    public static final double FACING_AWAY_FROM_DRIVERS = 90;
+    public static final double FACING_TOWARDS_DRIVERS = -90;
+    
+    private final DoubleProperty currentPitch;
+    private final DoubleProperty currentRoll;
+    
+    private final DoubleProperty inherentRioPitch;
+    private final DoubleProperty inherentRioRoll;
+    
+    private double previousLeftDistance;
+    private double previousRightDistance;
+    
+    private BooleanProperty rioRotated;
+    
+    public BasePoseSubsystem(WPIFactory factory, XPropertyManager propManager) {
+        log.info("Creating");
+        imu = factory.getGyro(ImuType.navX);
+        
+        currentHeadingProp = propManager.createEphemeralProperty("CurrentHeading", 0.0);
+        // Right when the system is initialized, we need to have the old value be
+        // the same as the current value, to avoid any sudden changes later
+        lastImuHeading = imu.getYaw();
+        currentHeading = new ContiguousHeading(FACING_AWAY_FROM_DRIVERS);
+        
+        currentPitch = propManager.createEphemeralProperty("CurrentPitch", 0.0);
+        currentRoll = propManager.createEphemeralProperty("CurrentRoll", 0.0);
+        
+        leftDriveDistance = propManager.createEphemeralProperty("LeftDriveDistance", 0.0);
+        rightDriveDistance = propManager.createEphemeralProperty("RightDriveDistance", 0.0);
+        
+        totalDistanceX = propManager.createEphemeralProperty("TotalDistanceX", 0.0);
+        totalDistanceY = propManager.createEphemeralProperty("TotalDistanceY", 0.0);
+        
+        rioRotated = propManager.createPersistentProperty("RioRotated", false);
+        inherentRioPitch = propManager.createPersistentProperty("InherentRioPitch", 0.0);
+        inherentRioRoll = propManager.createPersistentProperty("InherentRioRoll", 0.0);
+    }
+    
+    private void updateCurrentHeading() {
+        // Old heading - current heading gets the delta heading        
+        double imuDeltaYaw = lastImuHeading.difference(imu.getYaw());
+
+        // add the delta to our current
+        currentHeading.shiftValue(imuDeltaYaw);
+        
+        // update the "old" value
+        lastImuHeading = imu.getYaw();
+        
+        currentHeadingProp.set(currentHeading.getValue());
+        
+        currentPitch.set(getRobotPitch());
+        currentRoll.set(getRobotRoll());
+    }  
+    
+    private void updateDistanceTraveled(double currentLeftDistance, double currentRightDistance) {
+       
+        leftDriveDistance.set(currentLeftDistance);
+        rightDriveDistance.set(currentRightDistance);
+        
+        double deltaLeft = currentLeftDistance - previousLeftDistance;
+        double deltaRight = currentRightDistance - previousRightDistance;
+        
+        double totalDistance = (deltaLeft + deltaRight) / 2;
+        
+        // get X and Y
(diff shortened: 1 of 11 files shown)


src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+    
+    private double getRealPitch() {
+        if (rioRotated.get()) {
+            return imu.getRoll();
+        }
+        return imu.getPitch();
+    }
+    
+    private double getRealRoll() {
```
WasabiFan: "real" is used to indicate that the result isn't corrected for the inherent roll and pitch, but this method adjusts it for the third dimension. That needs clarification.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        currentHeadingProp = propManager.createEphemeralProperty("CurrentHeading", 0.0);
+        // Right when the system is initialized, we need to have the old value be
+        // the same as the current value, to avoid any sudden changes later
+        lastImuHeading = imu.getYaw();
+        currentHeading = new ContiguousHeading(FACING_AWAY_FROM_DRIVERS);
+        
+        currentPitch = propManager.createEphemeralProperty("CurrentPitch", 0.0);
+        currentRoll = propManager.createEphemeralProperty("CurrentRoll", 0.0);
+        
+        leftDriveDistance = propManager.createEphemeralProperty("LeftDriveDistance", 0.0);
```
aschokking: We should have a pow wow to talk about property naming scheme. I agree we should strive for consistency but I think we might have different views on what that scheme should be.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        // update the "old" value
+        lastImuHeading = imu.getYaw();
+        
+        currentHeadingProp.set(currentHeading.getValue());
+        
+        currentPitch.set(getRobotPitch());
+        currentRoll.set(getRobotRoll());
+    }  
+    
+    private void updateDistanceTraveled(double currentLeftDistance, double currentRightDistance) {
```
aschokking: It's a tricky thing, all of those 3 terms can be interpreted through the 'total/absolute' lens. I think in this case the 'Traveled' qualifier is the key thing that helps clarify that this isn't an absolute position.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        leftDriveDistance.set(currentLeftDistance);
+        rightDriveDistance.set(currentRightDistance);
+        
+        double deltaLeft = currentLeftDistance - previousLeftDistance;
+        double deltaRight = currentRightDistance - previousRightDistance;
+        
+        double totalDistance = (deltaLeft + deltaRight) / 2;
+        
+        // get X and Y
+        double deltaY = Math.sin(currentHeading.getValue() * Math.PI / 180) * totalDistance;
```
aschokking: sounds good, this stuff is easy to fat finger.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        return getTravelVector().rotate(-(currentHeading.getValue() - 90)).clone();
+    }
+    
+    public void resetDistanceTraveled() {
+        totalDistanceX.set(0);
+        totalDistanceY.set(0);
+    }
+    
+    public void setCurrentHeading(double headingInDegrees){
+        currentHeading.setValue(headingInDegrees);
```
aschokking: The heading tracking is based off of deltas of gyro readings applied to this value so I think it should be okay?

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+    
+    /**
+     * This should be called as often as reasonably possible, to increase accuracy
+     * of the "distance traveled" calculation.
+     * 
+     * The PoseSubsystem can't directly own CANTalons, so some command will need to feed in the
```
aschokking: minor comment tweak, rather than discussing Talons here maybe it's just the drive positional sensors?

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+    
+    public void calibrateInherentRioOrientation() {
+        inherentRioPitch.set(getRealPitch());
+        inherentRioRoll.set(getRealRoll());
+    }
+    
+    @Override
+    public void updatePeriodicData() {
+        updatePose();
```
aschokking: I was just wondering if updatePose should ever be called outside of updatePeriodicData(), maybe just have the latter?

src/xbot/common/subsystems/pose/commands/SetRobotHeadingCommand.java
```diff
@@ -0,0 +1,35 @@
...
+    BasePoseSubsystem poseSubsystem;
+    double heading;
+    
+    @Inject
+    public SetRobotHeadingCommand(BasePoseSubsystem poseSubsystem) {
+        this.poseSubsystem = poseSubsystem;
+        heading = BasePoseSubsystem.FACING_AWAY_FROM_DRIVERS;
+    }
+    
+    public void setHeadingToApply(double heading) {
```
aschokking: I guess we would mostly only ever do this with 90 so no need to support a property based version as well? I suppose we could muck with the currentHeading directly if need be..

tests/xbot/common/subsystems/pose/TestPoseSubsystem.java
```diff
@@ -30,4 +35,10 @@ protected double getLeftDriveDistance() {
     protected double getRightDriveDistance() {
         return right.getPosition();
     }
+    
+    public void setDistanceTraveled(double left, double right) {
```
aschokking: this isn't exactly setting the 'distanceTraveled' directly though right? Might be about setDriveEncoderDistances ?

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        inherentRioPitch = propManager.createPersistentProperty("InherentRioPitch", 0.0);
+        inherentRioRoll = propManager.createPersistentProperty("InherentRioRoll", 0.0);
+    }
+    
+    private void updateCurrentHeading() {
+        // Old heading - current heading gets the delta heading        
+        double imuDeltaYaw = lastImuHeading.difference(imu.getYaw());
+
+        // add the delta to our current
+        currentHeading.shiftValue(imuDeltaYaw);
```
JohnGilb: This approach makes the setCurrentHeading method really simple.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
JohnGilb: went more general - `updateOdometry`

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
JohnGilb: changed `real` to `untrimmed`, as in https://en.wikipedia.org/wiki/Trim_tab

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
WasabiFan: Whatever you call this method, you should normalize to one of the two suggested names over "distance travelled".

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
WasabiFan: But is it any better than storing an offset? Doing so would mean that the "set" method saves the difference of the IMU's yaw and the given heading, and then the lines here would be simplified from three to one; a wash or even net benefit. That method also removes the unnatural scheme of calculating deltas just to convert it back to absolute.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
WasabiFan: Ah, I conflated `lastImuHeading` and `currentHeading`. Yes, this logic looks like it should be OK. However, see my comment on the core logic for the rotation above.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
JohnGilb: Sounds good

src/xbot/common/subsystems/pose/commands/SetRobotHeadingCommand.java
JohnGilb: Yeah - this is primarily used for big static starting positions, so they don't need to be dynamically changed too much at runtime.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
JohnGilb: The other advantage of this approach is that it lets the PoseSubsystem work with any given initial raw gyro value - no matter what value the gyro is at when the robot boots, since we're only looking at deltas in the measured value, we always start in a known config.

PR #70, opened 2017-02-19
Exposed and created rotational velocity in the Gyro interface. BasePo…

Unagifan:
XGyro and the classes which implement it now have the ability to see rotational velocity

### DIFF
--- a/src/xbot/common/controls/sensors/MockGyro.java
+++ b/src/xbot/common/controls/sensors/MockGyro.java
@@ -1,11 +1,13 @@
 package xbot.common.controls.sensors;
 
 import xbot.common.controls.MockRobotIO;
+import xbot.common.controls.sensors.navx.AHRS;
 import xbot.common.math.ContiguousHeading;
 
 public class MockGyro implements XGyro {
     private boolean isBroken;
     private MockRobotIO mockIO;
+    AHRS ahrs;
 
     public MockGyro(MockRobotIO mockRobotIO) {
         mockIO = mockRobotIO;
@@ -40,6 +42,10 @@ public class MockGyro implements XGyro {
     @Override
     public double getPitch() {
         return mockIO.getGyroPitch();
+    }
+    
+    public double getVelocityOfYaw(){
+        return ahrs.getRate();
     }
 
 }
--- a/src/xbot/common/controls/sensors/MockGyro.java
+++ b/src/xbot/common/controls/sensors/MockGyro.java
@@ -7,7 +7,6 @@ import xbot.common.math.ContiguousHeading;
 public class MockGyro implements XGyro {
     private boolean isBroken;
     private MockRobotIO mockIO;
-    AHRS ahrs;
 
     public MockGyro(MockRobotIO mockRobotIO) {
         mockIO = mockRobotIO;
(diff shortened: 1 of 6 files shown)


src/xbot/common/controls/sensors/MockGyro.java
```diff
@@ -1,11 +1,13 @@
 package xbot.common.controls.sensors;
 
 import xbot.common.controls.MockRobotIO;
+import xbot.common.controls.sensors.navx.AHRS;
```
WasabiFan: Why does the mock use the real class? Is this correct?

src/xbot/common/controls/sensors/MockGyro.java
```diff
@@ -41,5 +43,9 @@ public double getRoll() {
     public double getPitch() {
         return mockIO.getGyroPitch();
     }
+    
+    public double getVelocityOfYaw(){
```
WasabiFan: "angular velocity" is the typical phrase for "velocity of yaw".

src/xbot/common/controls/sensors/nav6/Nav6Gyro.java
```diff
@@ -16,6 +17,7 @@
     private SerialPort imuPort;
     private IMU coreSensor;
     private ContiguousHeading yawValue;
+    AHRS ahrs;
```
WasabiFan: Why are there so many instances of this? What _is_ this class?

src/xbot/common/controls/sensors/adapters/InertialMeasurementUnitAdapter.java
```diff
@@ -58,12 +58,15 @@ public double getRoll() {
...
 
     @Override
     public boolean isBroken() {
         return isBroken;
     }
-
-
+    
+    public double getVelocityOfYaw(){
+        return ahrs.getRate();
```
WasabiFan: What are the units? They should be documented.

src/xbot/common/controls/sensors/MockGyro.java
JohnGilb: Andy, the AHRS (attitude and heading reference system) should only be used in the InertialMeasurementUnitAdapter class.

In this mock class, you have to "mock" (a.ka. "fake") the behavior of the AHRS.

src/xbot/common/controls/sensors/MockGyro.java
JohnGilb: yeah, something like `getYawAngularVelocity()` would be clearer.

src/xbot/common/controls/sensors/nav6/Nav6Gyro.java
JohnGilb: Andy, the nav6 gyro is represented by the `IMU coreSensor` above, in the same way the NavX is represented by `AHRS`.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -14,6 +15,7 @@
 public abstract class BasePoseSubsystem extends BaseSubsystem implements PeriodicDataSource {
 
     public final XGyro imu;
+    AHRS ahrs;
```
JohnGilb: You already have the `public final XGyro imu`. Since you added the angular velocity method in the XGyro interface, you can just call that, rather than creating another AHRS and calling it.

WasabiFan:
If I tilt the NavX so that the previously horizontal plane is now vertical, does yaw continue to report the same value relative to the world? In other words, is the yaw value gravity- or sensor-relative?

JohnGilb:
I believe it's basically sensor-relative, but we can reconfigure to the gravity axis using omnimount: http://www.pdocs.kauailabs.com/navx-mxp/installation/omnimount/

PR #73, opened 2017-02-20
Feature/isontarget pid logic

JohnGilb:
Mostly written by Seth! =]

### DIFF
--- a/src/xbot/common/math/PID.java
+++ b/src/xbot/common/math/PID.java
@@ -26,13 +26,13 @@ public class PID
     
     private boolean checkErrorThreshold = false;
     private boolean checkDerivativeThreshold = false;
-    private boolean checkOnTargetForDuration = false;
+    private boolean checkTimeThreshold = false;
     private boolean waitingToStabilize = false;
-    private double onTargetThreshold = 0;
+    private double timeToleranceInSeconds = 0;
     
     private double startTime = 0;
     
-    private boolean onTarget = false;
+    private boolean onTargetStartTime = false;
     
     
     /**
@@ -66,10 +66,10 @@ public class PID
      *  e.g. if you wanted a minimum rotation speed of 5 degrees per second,
      *  this tolerance would need to be 0.25.  
      */
-    public void setTolerances(double errorTolerance, double derivativeTolerance, double timeTolerenceInSeconds) {
+    public void setTolerances(double errorTolerance, double derivativeTolerance, double timeToleranceInSeconds) {
         this.errorTolerance = errorTolerance;
         this.derivativeTolerance = derivativeTolerance;
-        this.onTargetThreshold = timeTolerenceInSeconds;
+        this.timeToleranceInSeconds = timeToleranceInSeconds;
     }
     
     /**
@@ -78,7 +78,7 @@ public class PID
     public void setShouldCheckTolerances(boolean checkError, boolean checkDerivative, boolean checkTime) {
         checkErrorThreshold = checkError;
         checkDerivativeThreshold = checkDerivative;
-        checkOnTargetForDuration = checkTime;
+        checkTimeThreshold = checkTime;
     }
     
     /**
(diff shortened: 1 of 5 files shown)


src/xbot/common/math/PID.java
```diff
@@ -55,22 +66,24 @@ public void reset()
      *  e.g. if you wanted a minimum rotation speed of 5 degrees per second,
      *  this tolerance would need to be 0.25.  
      */
-    public void setTolerances(double errorTolerance, double derivativeTolerance) {
+    public void setTolerances(double errorTolerance, double derivativeTolerance, double timeTolerenceInSeconds) {
```
aschokking: suggestion: timeWithinToleranceInSeconds

aschokking: or the onTargetDuration like the boolean name below?

src/xbot/common/math/PID.java
```diff
@@ -23,6 +26,14 @@
     
     private boolean checkErrorThreshold = false;
     private boolean checkDerivativeThreshold = false;
+    private boolean checkOnTargetForDuration = false;
+    private boolean waitingToStabilize = false;
+    private double onTargetThreshold = 0;
+    
+    private double startTime = 0;
```
aschokking: startTime is a little generic, onTargetStartTime maybe?

src/xbot/common/math/PID.java
```diff
@@ -151,27 +166,51 @@ public double calculate(double goal, double current,
...
         
-        return isOnTarget;
+        if (checkOnTargetForDuration) {
+            if(onTarget){
+                if(waitingToStabilize == false){
+                    waitingToStabilize = true;
+                    startTime = Timer.getFPGATimestamp();
+                    onTarget = false;
+                } else {
+                    onTarget =  Timer.getFPGATimestamp() - startTime > onTargetThreshold;
```
aschokking: is the fpgatimestamp in seconds? For some reason I thought it was ms

src/xbot/common/math/PIDPropertyManager.java
```diff
@@ -129,4 +146,12 @@ public boolean getEnableDerivativeThreshold() {
     public void setEnableDerivativeThreshold(boolean isEnabled) {
         propEnableDerivativeThreshold.set(isEnabled);
     }
+    
+    public boolean getEnableTimeThreshold() {
+        return propEnableTimeThreshold.get();
```
aschokking: I was just wondering if this was a thing we really needed enable/disable around or if a threshold of 0 is pretty self explanatory in this case?

src/xbot/common/math/PID.java
JohnGilb: Yeah, the naming's a bit weird, since it's a tolerance about tolerances. I'll think about this some more.

src/xbot/common/math/PID.java
JohnGilb: Sounds good

src/xbot/common/math/PID.java
JohnGilb: Yeah, I checked.

From WPILib's Timer.class:

/**
   * Return the system clock time in seconds. Return the time from the FPGA hardware clock in
   * seconds since the FPGA started.
   *
   * @return Robot running time in seconds.
   */

JohnGilb: I think years ago it was microseconds or something.

aschokking: Good to know!

src/xbot/common/math/PIDPropertyManager.java
JohnGilb: The enable/disable is more to support if you want to temporarily disable threshold checking, but want to keep your configured threshold.

aschokking: Ahh gotcha, that makes sense. Is that a thing we think we might do?

JohnGilb: Honestly, I'm not sure. I initially added the enable/disable for the other tolerances thinking we might, but we haven't so far.

On the other hand, we haven't done a lot of PID tuning yet, so I'm thinking we should keep these around for a little bit longer to see if we use them at all.

PR #74, opened 2017-02-20
Move XboxController logic into a base class

JohnGilb:
This moves all our extra gamepad logic (like inversion) into a base abstract class, and creates a "real" and "mock" implementation that fulfill that abstract class.

The advantage of this approach is that all our extra logic is now in a single place and testable, rather than having to be duplicated in the mock class.


src/xbot/common/controls/sensors/BaseXboxControllerAdapter.java
```diff
@@ -0,0 +1,211 @@
+package xbot.common.controls.sensors;
```
aschokking: having the name Base and Adapter is kind of rough, maybe just `XboxControllerAdapter`

JohnGilb: maybe go with the old style of `XXboxController` ?

PR #75, opened 2017-02-20
instead of implementing every possible method, let's only implement t…

JohnGilb:
…he ones we use, and protect most of them.

### DIFF
--- a/src/xbot/common/controls/sensors/MockXboxControllerAdapter.java
+++ b/src/xbot/common/controls/sensors/MockXboxControllerAdapter.java
@@ -1,8 +1,6 @@
 package xbot.common.controls.sensors;
 
-import edu.wpi.first.wpilibj.GenericHID.HIDType;
 import edu.wpi.first.wpilibj.GenericHID.Hand;
-import edu.wpi.first.wpilibj.GenericHID.RumbleType;
 import xbot.common.math.XYPair;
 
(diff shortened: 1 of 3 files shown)


src/xbot/common/controls/sensors/MockXboxControllerAdapter.java
```diff
@@ -27,192 +25,36 @@ public MockXboxControllerAdapter(int port) {
     }
 
     @Override
-    public String toString() {
-        // TODO Auto-generated method stub
-        return null;
-    }
-
-    @Override
-    public void setRumble(RumbleType type, double value) {
```
WasabiFan: This API should be kept; it's core functionality that we can take advantage of.

src/xbot/common/controls/sensors/MockXboxControllerAdapter.java
```diff
@@ -27,192 +25,36 @@ public MockXboxControllerAdapter(int port) {
...
-    @Override
-    public boolean getStartButton() {
-        // TODO Auto-generated method stub
-        return false;
-    }
-
-    @Override
-    public boolean getBackButton() {
+    protected boolean getRawButton(int button) {
         // TODO Auto-generated method stub
```
WasabiFan: There should be no TODOs in this file

PR #76, opened 2017-02-22
added new method to get the data for influx

imathno:
This is required for the travis to pass on the other influx branch


commonlib-build.properties
```diff
@@ -18,13 +18,14 @@ inject.jar=lib/guice-3.0/javax.inject.jar
 log4j.jar=lib/log4j/log4j-1.2.17.jar
 commonsio.jar=lib/commons-io-2.4/commons-io-2.4.jar
 derby.jar=lib/Derby/derby.jar
+influx.jar=lib/influxdb-java-2.5.jar
```
aschokking: It seems like if we go this route we need put all of the jars here (and move them into the commonlib/lib folder)

imathno: Is there a way for the commonlib to use jars in competition's lib?

aschokking: Good question, fundamentally no, and the reason is that the common lib is in theory common, it shouldn't depend on the competition project at all. For instance the common lib was used in the xbotedu project you were using before.

PR #77, opened 2017-02-24
Prevent error when property's key is missing from the store

WasabiFan:
There's a missing check in the property classes which opens us up to fatal errors (to the extent of "Robots don't quit!") when the underlying data store is missing the property's key. This PR fixes that issue and replaces the `NullPointerException` with an angry error print.

Note that I decided to make the `StringProperty` match the others even though it isn't strictly necessary; it seems better to return the default value instead of `null`.

### DIFF
--- a/src/xbot/common/properties/BooleanProperty.java
+++ b/src/xbot/common/properties/BooleanProperty.java
@@ -6,6 +6,7 @@
 
 package xbot.common.properties;
 
+import org.apache.log4j.Logger;
 
 /**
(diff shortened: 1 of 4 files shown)


src/xbot/common/properties/BooleanProperty.java
```diff
@@ -6,6 +6,7 @@
 
 package xbot.common.properties;
 
+import org.apache.log4j.Logger;
```
aschokking: is this import required?

PR #78, opened 2017-02-24
Log the new heading in SetRobotHeadingCommand

WasabiFan:


WasabiFan:
Good point; updated.

src/xbot/common/subsystems/pose/commands/SetRobotHeadingCommand.java
```diff
@@ -23,6 +23,7 @@ public void setHeadingToApply(double heading) {
     @Override
     public void initialize() {
         log.info("Initializing");
+        log.info("Setting heading to " + heading);
```
aschokking: Might consider merging these 2 log lines to be a little more concise

WasabiFan: With a newline character in the middle? I figure it's easier to read as two separate lines than as one concatenated string.

PR #80, opened 2017-03-03
increased storage space for type and value

tfrancisco:


aschokking:
Thanks for doing this, 100 for the value sound great!
I was curious, what `type` did we have was longer than 20 chars or was it more why not?

WasabiFan:
Now that this has been merged, we must go through each RIO and re-create the database. @JohnGilb @aschokking Who's going to take point on that?

aschokking:
Created a task: https://app.asana.com/0/38541457243752/286018733793663

WasabiFan:
We need to make sure that this is done today. If we have updated our code, we _must_  make sure that all databases match immediately. Otherwise, we will not have any hope of tracking which have been modified and which haven't, and code that works on one will throw difficult-to-detect errors which cause our robot to operate unexpectedly.

aschokking:
@tfrancisco has picked up the task at any rate

PR #81, opened 2017-03-04
Add FieldPose and some accessory methods

JohnGilb:
This is meant to be used as part of a command that will drive to arbitrary points on the field in a somewhat smooth manner.


src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
...
+        return deltaY/deltaX;
+    }
+    
+    private double slopeToDegrees(double slope) {
+        double x = 1;
+        if (Math.abs(this.m) > 90) {
+            x = -1;
+        }
+        
+        double rads = Math.atan2(this.m, x);
```
WasabiFan: `atan2` operates on X- and Y-coordinates... these parameters are slope and X-coordinate. What does this do?

src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
...
+    
+    public FieldPose getLinePerpendicularToPoint(XYPair point) {
+        // for lines of zero slope, this is not great
+        
+        double perp_m = 0;
+        
+        if (m == 0) {
+            // if we have no slope, instead of a line with infinite slope, we just use a very large slope, 
+            // like a million.
+            perp_m = 1000000;
```
WasabiFan: 😟

src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private double m;
+    private double b;
+    private ContiguousHeading heading;
+    private final XYPair fieldPosition;
+    
+    public FieldPose(XYPair point, ContiguousHeading heading) {
+        m = degreesToSlope(heading.getValue());
+        this.fieldPosition = point;
+        this.heading = heading;
```
WasabiFan: `.clone` these so the caller can't modify them after they're passed in

src/xbot/common/math/XYPair.java
```diff
@@ -79,6 +79,10 @@ public XYPair add(XYPair pair) {
         this.y += pair.y;
         return this;
     }
+    
+    public double getDistanceToPoint(XYPair otherPoint) {
+        return Math.sqrt(Math.pow((otherPoint.x - this.x), 2) + Math.pow((otherPoint.y - this.y), 2));
```
WasabiFan: This could be made more concise by either subtracting the vectors and getting the magnitude or using `Math.hypot`

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -1,5 +1,7 @@
 package xbot.common.subsystems.pose;
 
+import java.text.FieldPosition;
```
WasabiFan: 😨 🚨 ⏰ 💣 💥

src/xbot/common/math/ContiguousHeading.java
```diff
@@ -15,4 +15,11 @@ public ContiguousHeading clone() {
         return new ContiguousHeading(this.getValue());
     }
     
+    @Override
+    public ContiguousHeading shiftValue(double shiftMagnitude) {
+        // TODO Auto-generated method stub
```
JohnGilb: Good catch, thanks

src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
+package xbot.common.math;
+
+public class FieldPose {
+
+    private double m;
```
JohnGilb: Yes, that can be made more explicit.

src/xbot/common/math/FieldPose.java
JohnGilb: Nice catch!

src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
...
+    private ContiguousHeading heading;
+    private final XYPair fieldPosition;
+    
+    public FieldPose(XYPair point, ContiguousHeading heading) {
+        m = degreesToSlope(heading.getValue());
+        this.fieldPosition = point;
+        this.heading = heading;
+        
+        // if we already have the y-intercept, avoid division by 0.
+        if ((point.x == 0) || (Math.abs(m) < 0.01)) {
```
JohnGilb: Yes, a RobotPose is a combination of a point/heading, as well as the line described by that heading (so that a robot could attempt to get on that line of travel)

src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
...
+            b = point.y;
+        } else {
+            b = point.y / (point.x * m); 
+        }   
+    }
+    
+    public FieldPose(XYPair point, double slope, boolean positiveAngle) {
+        this.fieldPosition = point;
+        m = slope;
+        b = point.y - (point.x * m);
```
JohnGilb: Oh man, they are both supposed to be subtract... means my tests aren't very good if they missed this!

JohnGilb: I've added a test that fails under the old logic, and passes with the fix.

src/xbot/common/math/FieldPose.java
JohnGilb: A slope is just "how much does y change for a change of 1 in x", so you can essentially create coordinates by setting y = slope and x = 1.

src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
...
+        if (Math.abs(this.m) > 90) {
+            x = -1;
+        }
+        
+        double rads = Math.atan2(this.m, x);
+        
+        return Math.toDegrees(rads);
+    }
+    
+    public FieldPose getLinePerpendicularToPoint(XYPair point) {
```
JohnGilb: Haha, yes, this is a pretty terrible name. Maybe `getPerpendicularLineThatIncludesPoint()` would be clearer.

src/xbot/common/math/FieldPose.java
WasabiFan: Oh, I get it; I missed the fact that this was just `1` or `-1` and was thinking it was the x-coordinate of a point on the line.

src/xbot/common/math/FieldPose.java
WasabiFan: What is it that using a slope gives you that a heading does not? I'm trying to understand why we need to keep track of slope and its special cases while seemingly only using the heading in consumer code.

src/xbot/common/math/FieldPose.java
WasabiFan: `getPerpendicularLineCoincidentWithPoint`?

src/xbot/common/math/FieldPose.java
JohnGilb: Yeah, I didn't want to deal with Inf and Nan, and got kinda lazy.

src/xbot/common/math/FieldPose.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+    
+    public double getY(double x) {
+        return x*m+b;
+    }
+    
+    private XYPair getIntersectionWithLine(FieldPose line) {
+        // calculate X point where they meet
+        
+        double x_intersect = (line.b - this.b) / (this.m - line.m);
```
JohnGilb: When two lines meet, it's appropriate to call that an "intersection"
http://math.stackexchange.com/questions/216151/difference-between-intercept-and-intersect

src/xbot/common/math/FieldPose.java
WasabiFan: This is one of those special cases that makes me confused about why we're using slope instead of heading.

src/xbot/common/math/FieldPose.java
WasabiFan: Ah, so this is the coordinate of the intersection; could you call this something like `intersectionX`?

JohnGilb: Added some comments. Basically, this code calculates the x-intercept of two lines, and uses that to find the y-intercept.

src/xbot/common/math/XYPair.java
JohnGilb: Cleaned this up a bit, used existing methods.

src/xbot/common/subsystems/pose/BasePoseSubsystem.java
JohnGilb: 🚨 !!

src/xbot/common/math/FieldPose.java
JohnGilb: I need the slope so I can have a line - because it's easy to calculate the intersection of two lines.

PR #82, opened 2017-03-04
Feature/reset heading and distance

SethNguyenXBOT:


src/xbot/common/subsystems/pose/ResetHeadingAndDistanceCommandGroup.java
```diff
@@ -0,0 +1,23 @@
...
+    @Inject
+    public ResetHeadingAndDistanceCommandGroup(BasePoseSubsystem poseSubsystem,
+            ResetDistanceCommand resetDistanceCommand,
+            SetRobotHeadingCommand setRobotHeadingCommand) {
+        ResetDistanceCommand resetDistance = 
+                new ResetDistanceCommand(poseSubsystem);
+        SetRobotHeadingCommand resetHeading =
+                new SetRobotHeadingCommand(poseSubsystem);
+        
+        this.addParallel(resetDistance);
```
WasabiFan: These both require the pose subsystem, right? What happens when two commands which require the same subsystem are run in parallel?

aschokking: CommandGroup's collect all their subcommands requirements at the start and globally require them all for the duration of the command group

PR #83, opened 2017-03-06
Use parametric form of line for field pose

WasabiFan:
I re-wrote the `FieldPose` tests to test the core external APIs instead of the slope math (the primary interfaces weren't being tested) then replaced the `FieldPose` with my modified implementation using a parametric approach and confirmed that those tests -- plus the additional vertical case which previously failed -- pass.


JohnGilb:

WasabiFan:
If we aren't going to merge this, we should delete it entirely from the repository. At the moment the implementation is broken, so it's just a hazard. Either way, I don't see any reason not to merge this — all tests pass here which passed in the existing version, plus some new ones which didn't.

WasabiFan:
I updated the PR; do we even want this pose functionality though? I'm not sure how it figures into your plans.

PR #85, opened 2017-03-12
Feature/track xbox buttons

JohnGilb:


src/xbot/common/controls/sensors/XXboxController.java
```diff
@@ -48,6 +56,16 @@ public int getValue() {
     }
     
     public AdvancedXboxButton getXboxButton(XboxButton buttonName) {
+        
+        if (!allocatedButtons.containsKey(buttonName)) {
+            allocatedButtons.put(buttonName, true);
```
WasabiFan: Could we instead maintain a map of `AdvancedXboxButton`s which contains all the buttons we've returned and check against that? It feels odd to be returning a button which we lose all references to and instead maintain state that describes whether we've ever generated a given button. I'm thinking that keeping references to these buttons might be useful in the future.

tests/xbot/common/controls/sensors/XboxControllerTest.java
```diff
@@ -0,0 +1,27 @@
...
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+    
+    @Test
+    public void doubleAllocateButton() {
+        WPIFactory factory = injector.getInstance(WPIFactory.class);
+        XXboxController controller = factory.getXboxController(0);
+        
+        // We expect the robot to get mad if we try to get the same button twice
+        thrown.expect(RobotAssertionException.class);
```
WasabiFan: Is there an advantage to doing it this way as opposed to using the annotation?

JohnGilb: This will fail the test if the exception fails to throw - I think the annotation just says ignore it if it happens?

src/xbot/common/controls/sensors/XXboxController.java
JohnGilb: Good future-proofing.

tests/xbot/common/controls/sensors/XboxControllerTest.java
JohnGilb: Ah, nevermind, I see how it works now. Yep, that's cleaner.

I think this rule approach is meant when you want to enable/disable checking during the test itself.

PR #86, opened 2017-03-17
Add multiplexed LIDAR pair

WasabiFan:
Equivalent of this:

```Arduino
void tcaselect(uint8_t i) {
  if (i > 7) return;
 
  Wire.beginTransmission(TCAADDR);
  Wire.write(1 << i);
  Wire.endTransmission();  
}
```


WasabiFan:
We need to physically tie three pins to ground; other than that, there's no configuration required.

PR #88, opened 2017-03-26
Remove threaded distance sensor updater

WasabiFan:


WasabiFan:
PR updated; @JohnGilb can I get another review?

PR #89, opened 2017-03-26
Prevent NullPointerException on nonexistent property

WasabiFan:


src/xbot/common/properties/BooleanProperty.java
```diff
@@ -56,9 +56,9 @@ public void set(boolean value) {
      * Saves the value permanently (presumably to the Robot).
      */
     public void save() {
-        if(persistenceType == PropertyPersistenceType.Persistent) {
-            permanentStore.setBoolean(key, randomAccessStore.getBoolean(key)
-                .booleanValue());
+        Boolean value;
+        if(persistenceType == PropertyPersistenceType.Persistent && (value = randomAccessStore.getBoolean(key)) != null) {
+            permanentStore.setBoolean(key, value.booleanValue());
```
JohnGilb: `else` warn logging?

PR #90, opened 2017-03-27
Log match info and loop frequency in BaseRobot

WasabiFan:


src/xbot/common/command/BaseRobot.java
```diff
@@ -82,21 +88,37 @@ protected void initializeSystems() {
...
+            + (ds.isDSAttached() ? "DS connected" : "DS disconnected") + ", "
+            + (ds.isFMSAttached() ? "FMS connected" : "FMS disconnected") + ", "
+            + "Is disabled: " + ds.isDisabled() + ", "
+            + "Is enabled: " + ds.isEnabled() + ", "
+            + "Is auto: " + ds.isAutonomous() + ", "
+            + "Is teleop: " + ds.isOperatorControl() + ", "
+            + "Is test: " + ds.isTest() + ", "
+            + "Is browned out: " + ds.isBrownedOut() + ", "
+            + "Is output enabled: " + ds.isSysActive() + ", "
+            + "Battery voltage: " + ds.getBatteryVoltage();
```
imathno: Should you have a new line after each is statement?

WasabiFan: I'm keeping it on one line right now so you can search through the file and you only need to see the single line with the match to get the info.

WasabiFan:
@aschokking can I get your review on this one? I plan to merge and deploy this tomorrow before we go on the field.

src/xbot/common/command/BaseRobot.java
```diff
@@ -64,12 +68,14 @@ public void robotInit() {
             DOMConfigurator.configure("log4j.xml");
         } catch (Exception e) {
             // Had a problem loading the config. Robot should continue!
-            System.out.println("Couldn't configure logging - file probably missing or malformed");
+            log.error("Couldn't configure logging - file probably missing or malformed");
```
aschokking: I think using System.out here was intentional since the `log` might not go anywhere in this case. Maybe do both?

src/xbot/common/command/BaseRobot.java
```diff
@@ -132,6 +154,17 @@ public void testPeriodic() {
     protected void sharedPeriodic() {
         xScheduler.run();
         this.updatePeriodicDataSources();
+        
+        loopCycleCounter++;
+        double timeSinceLastLog = Timer.getFPGATimestamp() - lastFreqCounterResetTime;
+        if(lastFreqCounterResetTime > 0 && timeSinceLastLog >= frequencyReportInterval.get()) {
```
aschokking: I think `lastFreqCounterResetTime` starts as -1 and is only set to something larger inside this `if` so it would never actually execute?

PR #93, opened 2017-04-12
Add logging to BaseCommandGroup and log interrupted

aschokking:


src/xbot/common/command/BaseCommandGroup.java
```diff
@@ -1,16 +1,37 @@
...
     @Inject
     SmartDashboardCommandPutter commandPutter;
     
+    public BaseCommandGroup() {
+        log = Logger.getLogger(this.getName());
+    }
+
+    public BaseCommandGroup(String name) {
+        super(name);
+        log = Logger.getLogger(this.getName());
```
WasabiFan: I find this much easier to follow if you use the local `name`; otherwise I have to search up the tree to confirm that `super` eventually sets a variable which will allow this to work properly.

aschokking: Fair point, I can see it both ways. This way enables other manipulation/overriding of what the name shakes out to be in `getName()` and that's what will appear in the smartdashboard so it would be nice if they stayed in sync. On the other hand as you said you do have to spelunk through to verify.

This is currently how things are setup in BaseCommand so in the short-run I'd vote for consistency with this PR and if we want to change it up we can modify both.

PR #94, opened 2017-04-12
Feature/current and loop duration logging

Unagifan:
Implemented logging for current and loop duration

### DIFF
--- a/src/xbot/common/command/BaseRobot.java
+++ b/src/xbot/common/command/BaseRobot.java
@@ -51,6 +51,7 @@ public class BaseRobot extends IterativeRobot {
     protected DoubleProperty frequencyReportInterval;
     protected double lastFreqCounterResetTime = -1;
     protected int loopCycleCounter = 0;
+    protected double loopDuration = 0;
     
     protected ArrayList<PeriodicDataSource> periodicDataSources = new ArrayList<PeriodicDataSource>();
 
@@ -182,7 +183,7 @@ public class BaseRobot extends IterativeRobot {
         
         brownoutLatch.setValue(DriverStation.getInstance().isBrownedOut());
         
-        double loopDuration = Timer.getFPGATimestamp() - loopBegin;
+        loopDuration += Timer.getFPGATimestamp() - loopBegin;
         
         loopCycleCounter++;
         double timeSinceLastLog = Timer.getFPGATimestamp() - lastFreqCounterResetTime;
@@ -191,7 +192,9 @@ public class BaseRobot extends IterativeRobot {
         }
         else if(timeSinceLastLog >= frequencyReportInterval.get()) {
             double loopsPerSecond = loopCycleCounter / timeSinceLastLog; 
+            double averageLoopDuration = loopDuration / loopCycleCounter;
             
+            loopDuration = 0;
             loopCycleCounter = 0;
             lastFreqCounterResetTime = Timer.getFPGATimestamp();
             
@@ -199,9 +202,8 @@ public class BaseRobot extends IterativeRobot {
             
             log.info("Total Current: " + pdp.getTotalCurrent());
             
-            log.info("Loop Duration:" + loopDuration);
-        }
-        
+            log.info("Average Loop Duration:" + loopDuration);
+        } 
     }
     
     protected void registerPeriodicDataSource(PeriodicDataSource telemetrySource) {
--- a/src/xbot/common/command/BaseRobot.java
+++ b/src/xbot/common/command/BaseRobot.java
@@ -176,12 +176,14 @@ public class BaseRobot extends IterativeRobot {
     }
     
     protected void sharedPeriodic() {
+        double loopBegin = Timer.getFPGATimestamp();
         xScheduler.run();
         this.updatePeriodicDataSources();
         
         brownoutLatch.setValue(DriverStation.getInstance().isBrownedOut());
         
-        double loopBegin = Timer.getFPGATimestamp();
+        double loopDuration = Timer.getFPGATimestamp() - loopBegin;
+        
         loopCycleCounter++;
         double timeSinceLastLog = Timer.getFPGATimestamp() - lastFreqCounterResetTime;
         if(lastFreqCounterResetTime <= 0) {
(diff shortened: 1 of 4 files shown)


src/xbot/common/command/BaseRobot.java
```diff
@@ -180,14 +182,18 @@ protected void sharedPeriodic() {
         double timeSinceLastLog = Timer.getFPGATimestamp() - lastFreqCounterResetTime;
         if(lastFreqCounterResetTime <= 0) {
             lastFreqCounterResetTime = Timer.getFPGATimestamp();
+            log.info("Total Current:" + pdp.getTotalCurrent());
         }
         else if(timeSinceLastLog >= frequencyReportInterval.get()) {
             double loopsPerSecond = loopCycleCounter / timeSinceLastLog; 
+            double loopDuration = pdp.getTotalCurrent() / loopCycleCounter;
```
aschokking: I don't think this means what you think it means.

src/xbot/common/command/BaseRobot.java
```diff
@@ -55,6 +57,7 @@ public BaseRobot() {
         super();
         setupInjectionModule();
         
+        pdp = injector.getInstance(PowerDistributionPanelWPIAdapter.class);
```
JohnGilb: This won't work here. It's gotta be in `initializeSystems` next to all the other things getting created.

src/xbot/common/command/BaseRobot.java
```diff
@@ -180,14 +183,18 @@ protected void sharedPeriodic() {
         double timeSinceLastLog = Timer.getFPGATimestamp() - lastFreqCounterResetTime;
         if(lastFreqCounterResetTime <= 0) {
             lastFreqCounterResetTime = Timer.getFPGATimestamp();
+            log.info("Total Current:" + pdp.getTotalCurrent());
         }
         else if(timeSinceLastLog >= frequencyReportInterval.get()) {
             double loopsPerSecond = loopCycleCounter / timeSinceLastLog; 
+            double loopDuration = timeSinceLastLog / loopCycleCounter;
```
JohnGilb: Noooooooo this is just the inverse!

You gotta actually keep track of two times - when the loop started, and when the loop finished.

src/xbot/common/command/BaseRobot.java
Unagifan: !

src/xbot/common/command/BaseRobot.java
```diff
@@ -180,14 +183,18 @@ protected void sharedPeriodic() {
         double timeSinceLastLog = Timer.getFPGATimestamp() - lastFreqCounterResetTime;
         if(lastFreqCounterResetTime <= 0) {
             lastFreqCounterResetTime = Timer.getFPGATimestamp();
+            log.info("Total Current:" + pdp.getTotalCurrent());
```
Unagifan: !

src/xbot/common/command/BaseRobot.java
```diff
@@ -37,6 +38,7 @@
 
     protected XPropertyManager propertyManager;
     protected XScheduler xScheduler;
+    protected PowerDistributionPanelWPIAdapter pdp;
```
Unagifan: !

src/xbot/common/command/BaseRobot.java
WasabiFan: This needs to be moved elsewhere; right now, this will only log once when the loop runs the first time.

src/xbot/common/controls/sensors/wpi_adapters/PowerDistributionPanelWPIAdapter.java
```diff
@@ -15,6 +15,15 @@ public double getCurrent(int channel) {
         return pdp.getCurrent(channel);
     }
     
+    @Override
+    public double getTotalCurrent(){
+        double totalCurrent = 0;
+        for(int i = 0; i <= 15; i++){
```
WasabiFan: Is there another way to get the number of channels? This seems sketchy.

src/xbot/common/command/BaseRobot.java
Unagifan: !

src/xbot/common/controls/sensors/wpi_adapters/PowerDistributionPanelWPIAdapter.java
Unagifan: gotchu fam

src/xbot/common/command/BaseRobot.java
```diff
@@ -188,7 +196,12 @@ else if(timeSinceLastLog >= frequencyReportInterval.get()) {
             lastFreqCounterResetTime = Timer.getFPGATimestamp();
             
             log.info("Robot loops per second: " + loopsPerSecond);
+            
+            log.info("Total Current: " + pdp.getTotalCurrent());
```
WasabiFan: Can we remove these empty lines?

src/xbot/common/command/BaseRobot.java
```diff
@@ -188,7 +196,12 @@ else if(timeSinceLastLog >= frequencyReportInterval.get()) {
...
+            
+            log.info("Loop Duration:" + loopDuration);
```
JohnGilb: So there's still a logic problem here - you only report the most recent loop duration. After all, this chunk of code is only called every 10 seconds or so, during which there should have been several hundred loops.

What you need to do, after calculating the loop duration above, is accumulate it somehow. Then, in this code, you take the accumulated value and divide it by the number of loops, thus measuring the "average loop duration."

Unagifan: !

src/xbot/common/command/BaseRobot.java
```diff
@@ -171,24 +177,33 @@ public void testPeriodic() {
...
+        loopDuration += Timer.getFPGATimestamp() - loopBegin;
+        
         loopCycleCounter++;
         double timeSinceLastLog = Timer.getFPGATimestamp() - lastFreqCounterResetTime;
         if(lastFreqCounterResetTime <= 0) {
             lastFreqCounterResetTime = Timer.getFPGATimestamp();
         }
         else if(timeSinceLastLog >= frequencyReportInterval.get()) {
             double loopsPerSecond = loopCycleCounter / timeSinceLastLog; 
+            double averageLoopDuration = loopDuration / loopCycleCounter;
```
JohnGilb: What happens if loopCycleCounter is zero?

JohnGilb:

PR #95, opened 2017-04-15
Add class which notifies caller when sytem delta is over threshold

imathno:
Some description provided.


src/xbot/common/logging/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,36 @@
...
+    
+    private double oldPower = 0;
+    
+    public PeriodicDeltaObserver(String callerName, String message, double checkPowerThreshold) {
+        this.callerName = callerName;
+        this.message = message;
+        this.checkPowerThreshold = checkPowerThreshold;
+    }
+    
+    public void setMessage(String message) {
```
WasabiFan: I think you should change this so that the class itself only supplies a Boolean which indicates whether the delta was passed, and the consumer does the logging.

src/xbot/common/logging/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class PeriodicDeltaObserver {
+
+    protected static Logger log = Logger.getLogger(PeriodicDeltaObserver.class);
+    private final String callerName;
+    private String message;
+    
+    private double checkPowerThreshold;
+    
+    private double oldPower = 0;
```
WasabiFan: These aren't necessarily powers

src/xbot/common/logging/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,36 @@
...
+    
+    public void setCheckPowerThreshold(double power) {
+        this.checkPowerThreshold = power;
+    }
+    
+    public void checkPower(double newPower) {
+        if (Math.abs(oldPower - newPower) > checkPowerThreshold) {
+            log.info("From " + callerName + ": " + message);
+        }
+        oldPower = newPower;
```
WasabiFan: If we didn't notify, we don't want to replace the power value. We want to notify every time the value has changed by more than the threshold _since the last time the caller was notified_.

src/xbot/common/logging/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,30 @@
...
+    public void setCheckPowerThreshold(double power) {
+        this.checkPowerThreshold = power;
+    }
+    
+    public boolean isThereASignficantChangeInTheSystem(double newPower) {
+        if (Math.abs(oldPower - newPower) > checkPowerThreshold) {
+            oldPower = newPower;
+            return true;
+        }
+        oldPower = newPower;
```
WasabiFan: If we didn't notify, we don't want to replace the power value. We want to notify every time the value has changed by more than the threshold _since the last time the caller was notified_.

src/xbot/common/logging/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.apache.log4j.Logger;
+
+public class PeriodicDeltaObserver {
+
+    protected static Logger log = Logger.getLogger(PeriodicDeltaObserver.class);
+    
+    private double checkPowerThreshold;
+    
+    private double oldPower = 0;
```
WasabiFan: These aren't necessarily powers.

aschokking: consider going with `value` instead of `power` since this is more generic than motor powers?

src/xbot/common/logic/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,30 @@
...
+    public PeriodicDeltaObserver(double checkValueThreshold) {
+        this.checkValueThreshold = checkValueThreshold;
+    }
+    
+    public void setCheckPowerThreshold(double value) {
+        this.checkValueThreshold = value;
+    }
+    
+    public boolean isDelta(double newValue) {
+        if (Math.abs(oldValue - newValue) > checkValueThreshold) {
```
JohnGilb: I would rewrite this to be

`boolean significantChange = Math.abs(oldValue - newValue) > checkValueThreshold;`
oldValue = newValue;
return significantChange;`

src/xbot/common/logic/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,24 @@
...
+public class PeriodicDeltaObserver {
+
+    private double checkValueThreshold;
+    private Double lastReportedValue = null;
+    
+    public PeriodicDeltaObserver(double checkValueThreshold) {
+        this.checkValueThreshold = checkValueThreshold;
+    }
+    
+    public void setCheckPowerThreshold(double value) {
```
WasabiFan: "power" should be removed as this isn't about power. `setDeltaThreshold`?

src/xbot/common/logic/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,24 @@
+package xbot.common.logic;
+
+public class PeriodicDeltaObserver {
+
+    private double checkValueThreshold;
```
WasabiFan: Rename to include "delta" or similar -- `deltaThreshold`?

src/xbot/common/logic/PeriodicDeltaObserver.java
```diff
@@ -0,0 +1,24 @@
+package xbot.common.logic;
+
+public class PeriodicDeltaObserver {
```
WasabiFan: This isn't "periodic" anymore

JohnGilb:

PR #97, opened 2017-07-22
Use WPILib's Preferences system to store Properties

JohnGilb:
Instead of using the Derby database, we use the Preferences system, which has a simple way of serializing things to disk on the robot.

### DIFF
--- a/src/xbot/common/injection/RobotModule.java
+++ b/src/xbot/common/injection/RobotModule.java
@@ -9,6 +9,7 @@ import xbot.common.logging.SilentRobotAssertionManager;
 import xbot.common.math.PIDFactory;
 import xbot.common.properties.ITableProxy;
 import xbot.common.properties.PermanentStorage;
+import xbot.common.properties.PreferenceStorage;
 import xbot.common.properties.RobotDatabaseStorage;
 import xbot.common.properties.SmartDashboardTableWrapper;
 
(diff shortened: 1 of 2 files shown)


aschokking:
Does this work off-robot at all? Was just curious if it could be tested

JohnGilb:
@aschokking Doesn't work off-robot since it's directly invoking WPILib - but I did test it on the Robox.

aschokking:
In theory directly invoking wpilib doesn't mean it can't work off robot
though if it's just doing standard Java file io stuff?

WasabiFan:
The reason that WPILib code wouldn't work in a testing environment would be that it is invoking a native implementation. If this functionality is common across the C++ and Java versions of WPILib, it's likely to be using such native code and won't work on a PC. You should be able to look it up in their source repo if interested.

aschokking:
Sounds reasonable, I always forget about how much native stuff they have going on in there. Will dig in deeper to the code if the curiosity becomes overpowering :)

WasabiFan:
For future reference, the `Preferences` implementation is here: https://github.com/wpilibsuite/allwpilib/blob/master/wpilibj/src/athena/java/edu/wpi/first/wpilibj/Preferences.java

It makes some calls into their HAL and uses the native ntcore implementation, so it likely won't respond well to being run on a PC.

PR #98, opened 2017-10-01
Add a generic interface for using CAN devices

WasabiFan:
Sending from the RIO to an external custom device is confirmed working. I haven't yet tested the read operation.

### DIFF
--- a/src/xbot/common/controls/misc/wpi_adapters/CANDeviceWPIAdapter.java
+++ b/src/xbot/common/controls/misc/wpi_adapters/CANDeviceWPIAdapter.java
@@ -1,0 +1,43 @@
+package xbot.common.controls.misc.wpi_adapters;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+import xbot.common.controls.misc.XCANDevice;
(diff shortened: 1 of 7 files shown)


src/xbot/common/controls/misc/wpi_adapters/CANDeviceWPIAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+    private final int inboundArbitrationId;
+    private final int outboundArbitrationId;
+    
+    public CANDeviceWPIAdapter(int inboundArbitrationId, int outboundArbitrationId) {
+        this.inboundArbitrationId = inboundArbitrationId;
+        this.outboundArbitrationId = outboundArbitrationId;
+    }
+    
+    @Override
+    public void send(byte[] data) {
```
JohnGilb: Data appears to never be sent

WasabiFan:
@JohnGilb If you get a moment, could you update this for the new factory/injection structure?

WasabiFan:
After working on the CAN code a bit more, I've decided that I want to approach this differently; I am going to be using a section of the arbitration ID as a "packet type", so having individual class instances for each arbid doesn't make sense.

PR #99, opened 2017-10-28
Move from WPIFactory to CommonLibFactory

JohnGilb:
This is a new approach to the CommonLib, mostly aimed at increasing testability and ease of use:

- Logic actuators/sensors (things that interface with WPILib or other external libraries) have all of their logic moved into an abstract class, and the WPIAdapters and MockAdapters are thin shims. This means all of our low-level logic should be testable; no business logic should exist in the adapter classes anymore.

- WPIFactory has been replaced by CommonLibFactory, a factory automatically generated by Guice that uses Assisted Injection. 

The new pattern for updating CommonLibFactory is as follows:
1) Decorate your new object with @AssistedInject in the constructor for any arguments you need
2) Create a method in CommonLibFactory.java with the same decorators
(If your object needs a WPI/Mock adapter, you need the following steps as well)
3) In RobotModule.java, bind your class to its real implementation
4) In UnitTestModule.java, bind your clas to its mock implementation


tests/edu/wpi/first/wpilibj/MockAnalogInput.java
```diff
@@ -1,18 +1,23 @@
...
-    public MockAnalogInput(int channel, MockRobotIO mockRobotIO) {
+    @Inject
+    public MockAnalogInput(
+            @Assisted("channel") int channel, 
+            MockRobotIO mockRobotIO) {
         this.mockRobotIO = mockRobotIO;
         this.channel = channel;
     }
 
-    @Deprecated
```
WasabiFan: Why is this no longer deprecated?

tests/xbot/common/injection/wpi_factories/TestCommonLibFactory.java
```diff
@@ -0,0 +1,31 @@
...
+        clf.createEncoder("asdf", 1, 2, 1);
+        clf.createDigitalInput(1);
+        clf.createAnalogInput(1);
+        clf.createXboxController(1);
+        clf.createSolenoid(1);
+        clf.createDigitalOutput(1);
+        clf.createServo(1);
+        clf.createSpeedController(1);
+        clf.createCANTalon(1);
+        clf.createGyro();
```
WasabiFan: It'd be interesting to see if we could write a checkstyle rule or similar which enforced that everything showed up hear and in both modules...

tests/edu/wpi/first/wpilibj/MockAnalogInput.java
JohnGilb: Huh, no idea. Fixed though.

PR #104, opened 2017-12-16
Feature/base drive subsystem

JohnGilb:
Adds a BaseDriveSubsystem that holds a lot of the year-over-year driving logic.

The basic idea is that when you create Talons that correspond to wheels, you also say how they respond to the three major types of drive commands:
- Translation in the +X direction
- Translation in the +Y direction
- Rotation

Each wheel sums up its inputs, and that summed value is sent to the Talon to make it drive. This approach should work for Tank Drive, Mecanum, and Octocanum chassis. (It won't work for 3-wheel holonomic, but let's cross that bridge when we come to it).


src/xbot/common/controls/actuators/XCANTalon.java
```diff
@@ -185,6 +185,10 @@ public void ensureTalonControlMode(TalonControlMode mode) {
     public abstract void configForwardLimitSwitchNormallyOpen(boolean normallyOpen);
     public abstract void configReverseLimitSwitchNormallyOpen(boolean normallyOpen);
     
+    // Current --------------------------------------------
+    public abstract void enableCurrentLimit(boolean enable);
+    public abstract void setCurrentLimit(int amps);
```
WasabiFan: Is this an integer all the way through?

src/xbot/common/subsystems/BaseDriveSubsystem.java
```diff
@@ -0,0 +1,142 @@
...
+        public double w;
+        
+        public MotionRegistration(double x, double y, double w) {
+            this.x = x;
+            this.y = y;
+            this.w = w;
+        }        
+        
+        public double calculateTotalImpact(double x, double y, double w) {
+            return this.x * x + this.y*y + this.w * w;
```
WasabiFan: what's this for? The values are different units, so it seems like this would be useless...

src/xbot/common/subsystems/BaseDriveSubsystem.java
```diff
@@ -0,0 +1,142 @@
...
+     * list of Talons and how they respond to drive inputs.
+     */
+    protected abstract Map<XCANTalon, MotionRegistration> getAllMasterTalons();
+    
+    /**
+     * Returns the total distance tracked by the encoder
+     * - On the LEFT side of the robot
+     * - Pointing in the direction of +Y travel
+     */
+    public abstract double getLeftTotalDistance();
```
WasabiFan: How do you envision these three operating for tank and mecanum? Is transverse always zero on a tank robot?

src/xbot/common/subsystems/BaseDriveSubsystem.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    /**
+     * Determine the largest commanded output for a given wheel for a given
+     * translation/rotation input. For example, you would see a maximum output of
+     * two in a Tank Drive system that's commanded to go foward at full speed AND
+     * rotate at full speed.
+     */
+    protected double getMaxOutput(XYPair translation, double rotation) {
+        return getAllMasterTalons().values().stream()
+        .map(mr -> mr.calculateTotalImpact(translation.x, translation.y, rotation))
```
WasabiFan: Give `mr` a more readable name -- perhaps `master`

src/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
```diff
@@ -0,0 +1,64 @@
...
+            XPropertyManager propMan,
+            BasePoseSubsystem pose) {
+        this.headingModule = headingModule;
+        this.pose = pose;
+        humanThreshold = propMan.createPersistentProperty("HeadingAssistModule - Human Threshold", 0.05);
+        coastTime = propMan.createPersistentProperty("Heading Assist Module - Coast Time", 0.5);
+        lastHumanInput = 0;
+    }
+    
+    public double calculateHeadingPower(double humanRotationalPower) {
```
WasabiFan: It would be helpful to have a documentation comment with a reference implementation of calculating the rotational power param

src/xbot/common/subsystems/drive/control_logic/HeadingModule.java
```diff
@@ -0,0 +1,66 @@
...
+        // and you want to go to -170 degrees, you could travel -340 degrees... or just +20. 
+        
+        // So, we perform our own error calculation here that takes that into account (thanks to the ContiguousDouble
+        // class, which is aware of such circular effects), and then feed that into a PID where
+        // Goal is 0 and Current is our error.
+        
+        // This will have a side-effect of reversing the power output of the system, so we need to remember
+        // at the very end to reverse it one more time to cancel that out.
+        targetHeading.setValue(desiredHeading);
+        double errorInDegrees = pose.getCurrentHeading().difference(targetHeading);
```
WasabiFan: You could eliminate the later negation by swapping the args to `difference` -- `targetHeading.difference(pose.getCurrentHeading())`

src/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
```diff
@@ -0,0 +1,64 @@
...
+        // Also, update a timestamp that says this happened recently
+        if (humanRotationalPower > humanThreshold.get()) {
+            inAutomaticMode = false;
+            lastHumanInput = Timer.getFPGATimestamp();
+            return humanRotationalPower;
+        }
+        
+        // If not under threshold, but too close to timestamp,
+        // "coast"
+        double timeSinceHumanInput = Timer.getFPGATimestamp() - lastHumanInput;
```
WasabiFan: How will this work for tests? If I want to test driving, do I now also have to mock the timer and update it in my test?

src/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
```diff
@@ -0,0 +1,64 @@
...
+        
+        if (timeSinceHumanInput < coastTime.get()) {
+            return 0;
+        }
+        
+        if (timeSinceHumanInput > coastTime.get() && !inAutomaticMode) {
```
WasabiFan: Can we combine these into a single if..else chain? The code below should be equivalent to the existing code if the `>` here were replaced with `>=` (which I assume was the intent anyway).

```java
if (timeSinceHumanInput < coastTime.get()) {
    return 0;
}
else if (inAutomaticMode) {
    return headingModule.calculateHeadingPower(desiredHeading);
}   
else {
    desiredHeading = pose.getCurrentHeading().getValue();
    inAutomaticMode = true;
    headingModule.reset();
    return 0;
}
```

tests/xbot/common/injection/wpi_factories/TestCommonLibFactory.java
```diff
@@ -27,5 +29,7 @@ public void makeOneOfEverything() {
         clf.createLidarLite(I2C.Port.kOnboard);
         clf.createAdvancedJoystickButton(j, 1);
         clf.createAnalogHIDButton(j, 1, -1, 1);
+        
+        clf.createHeadingModule(pf.createPIDManager("testo", 1, 0, 0));
```
WasabiFan: `testo`?

tests/xbot/common/subsystems/drive/HeadingAssistModuleTest.java
```diff
@@ -0,0 +1,123 @@
...
+import xbot.common.subsystems.drive.control_logic.HeadingAssistModule;
+import xbot.common.subsystems.drive.control_logic.HeadingModule;
+
+public class HeadingAssistModuleTest extends BaseWPITest {
+
+    HeadingAssistModule ham;
+    
+    @Override
+    public void setUp() {
+        // TODO Auto-generated method stub
```
WasabiFan: remove

src/xbot/common/controls/actuators/XCANTalon.java
JohnGilb: In the current CTRE libraries, their API only exposes an integer. It may be that under the hood they change the value - I haven't looked that deep.

I think we can go with this for now, and review their source/api next year when we review their new libraries.

src/xbot/common/subsystems/BaseDriveSubsystem.java
JohnGilb: I'll make this more clear in some comments - these are actually all in the same "unitless" units: -100% to 100%. 

As an example, let's look at at the front left mecanum wheel. It would be configured as follows:
* X response: 1
* Y response: 1
* Rotation response: -1

If it's then commanded to go forward, strafe right, and rotate right, it will attempt to spin at 300% power.

(We know this isn't possible, which is why we can optionally choose to normalize the outputs if we want to avoid saturating the controls).

src/xbot/common/subsystems/BaseDriveSubsystem.java
JohnGilb: The DriveSubsystem is in charge of whatever encoders are related to drive - sometimes that's gearbox encoders, and sometimes we have "follow" encoders.

If we only have gearbox encoders, transverse would always be 0 for a tank robot.

However, if we had follow-wheel encoders, we could install a transverse encoder, which would be able to detect our robot being pushed or slipping, and so we would have some transverse value.

src/xbot/common/subsystems/BaseDriveSubsystem.java
JohnGilb: It took me a minute to remember what MR was supposed to stand for, which is not a good sign. =]

"motion registration"

Fixing.

src/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
JohnGilb: Excellent point. Including.

src/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
JohnGilb: Only if the system that you are testing uses the HeadingAssistModule. For example, if somebody uses that module to create a "TankDriveWithHeadingAssistCommand", proper testing of that class would likely involve messing around with time.

JohnGilb: And we have a nice solution for time manipulation - you can see it being used in HeadingAssistModuleTest.

src/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
JohnGilb: I like that your code is a little tighter (and "else if" avoids potential bugs where a bug could have multiple states executing together), but I also like the way that the current code represents the state machine order - typical execution of your code is in 1, 3, 2 order.

I'm going to use else if, but keep the execution order the same - does that seem reasonable?

src/xbot/common/subsystems/drive/control_logic/HeadingModule.java
JohnGilb: Good point, saves a line.

tests/xbot/common/injection/wpi_factories/TestCommonLibFactory.java
JohnGilb: Hah, just some "foo" kind of variable.  I'll update to classic random variable naming convention. =]

tests/xbot/common/subsystems/drive/HeadingAssistModuleTest.java
JohnGilb: Thanks. I need to turn on the Eclipse TODO finder.

JohnGilb:
Added a classic "left power, right power" tank implementation so that we don't need to do conversion in commands.

src/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
WasabiFan: Oh, I totally missed that this was part of a larger condition chain. Your approach sounds good to me!

PR #105, opened 2017-12-24
Add a class that monitors if packets are being sent from the driver station and warns if it's not

imathno:
- Added ConnectionMonitor class:
  The purpose of the class is to have a subsystem that keeps checking if we're connected to the drive 
  station. We can add some logic some where it vibrates the Xbox controllers or something to warn the 
  drivers. Then we can log when we were disconnecting.

- Added ConnectionMonitor Tests

### DIFF
--- a/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
+++ b/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
@@ -1,0 +1,40 @@
+package xbot.common.subsystems;
+
+import edu.wpi.first.wpilibj.Timer;
+import xbot.common.command.BaseSubsystem;
+import xbot.common.logic.Latch;
+
+import java.util.Observable;
+import java.util.Observer;
+
+public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem implements Observer {
+
+    private double timeOut;
+    private final Latch connectionLatch = new Latch(true, Latch.EdgeType.FallingEdge);
+
+    private double lastPacketReceivedTimeStamp = Timer.getFPGATimestamp();
+    private double disconnectionTime = -1;
+
+    public BaseConnectionMonitorSubsystem() {
+        connectionLatch.addObserver(this);
+    }
+
+    public void setTimeOut(double timeOut) {
+        this.timeOut = timeOut;
+    }
+
+    public synchronized void setLastPacketReceivedTimeStamp(double currentTimeStamp) {
+        connectionLatch.setValue((currentTimeStamp - lastPacketReceivedTimeStamp) < timeOut);
+        this.lastPacketReceivedTimeStamp = currentTimeStamp;
+    }
+
+    public double getPreviousDisconnectionTime() {
+        return disconnectionTime;
+    }
+
+    @Override
+    public void update(Observable o, Object arg) {
+        log.warn("The DriverStation has been disconnected for greater than " + timeOut + " seconds");
+        disconnectionTime = lastPacketReceivedTimeStamp;
+    }
+}
--- a/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
+++ b/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
@@ -19,7 +19,7 @@ public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem imple
 
     public BaseConnectionMonitorSubsystem(XPropertyManager propertyManager) {
         log.info("Creating");
-        timeOut = propertyManager.createPersistentProperty("ConnectionMontiorTimeOut", 1.0);
+        timeOut = propertyManager.createPersistentProperty("ConnectionMontiorTimeOut Seconds", 1.0);
         connectionLatch.addObserver(this);
     }
 
--- a/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
+++ b/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
@@ -3,38 +3,37 @@ package xbot.common.subsystems;
 import edu.wpi.first.wpilibj.Timer;
 import xbot.common.command.BaseSubsystem;
 import xbot.common.logic.Latch;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
 
 import java.util.Observable;
 import java.util.Observer;
 
 public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem implements Observer {
 
-    private double timeOut;
+    private DoubleProperty timeOut;
     private final Latch connectionLatch = new Latch(true, Latch.EdgeType.FallingEdge);
 
     private double lastPacketReceivedTimeStamp = Timer.getFPGATimestamp();
-    private double disconnectionTime = -1;
+    private double previousDisconnectionTimeStamp = -1;
 
-    public BaseConnectionMonitorSubsystem() {
+    public BaseConnectionMonitorSubsystem(XPropertyManager propertyManager) {
+        timeOut = propertyManager.createPersistentProperty("ConnectionMontiorTimeOut", 1.0);
         connectionLatch.addObserver(this);
     }
 
-    public void setTimeOut(double timeOut) {
-        this.timeOut = timeOut;
-    }
-
     public synchronized void setLastPacketReceivedTimeStamp(double currentTimeStamp) {
-        connectionLatch.setValue((currentTimeStamp - lastPacketReceivedTimeStamp) < timeOut);
+        connectionLatch.setValue((currentTimeStamp - lastPacketReceivedTimeStamp) < timeOut.get());
         this.lastPacketReceivedTimeStamp = currentTimeStamp;
     }
 
     public double getPreviousDisconnectionTime() {
-        return disconnectionTime;
+        return previousDisconnectionTimeStamp;
     }
 
     @Override
     public void update(Observable o, Object arg) {
         log.warn("The DriverStation has been disconnected for greater than " + timeOut + " seconds");
-        disconnectionTime = lastPacketReceivedTimeStamp;
+        previousDisconnectionTimeStamp = lastPacketReceivedTimeStamp;
     }
 }
--- a/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
+++ b/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
@@ -18,6 +18,7 @@ public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem imple
     private double previousDisconnectionTimeStamp = -1;
 
     public BaseConnectionMonitorSubsystem(XPropertyManager propertyManager) {
+        log.info("Creating");
         timeOut = propertyManager.createPersistentProperty("ConnectionMontiorTimeOut", 1.0);
         connectionLatch.addObserver(this);
     }
--- a/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
+++ b/src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
@@ -14,8 +14,8 @@ public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem imple
     private DoubleProperty timeOut;
     private final Latch connectionLatch = new Latch(true, Latch.EdgeType.FallingEdge);
 
-    private double lastPacketReceivedTimeStamp = Timer.getFPGATimestamp();
-    private double previousDisconnectionTimeStamp = -1;
+    private double lastPacketReceivedTimestamp = Timer.getFPGATimestamp();
+    private double previousDisconnectionTimestamp = -1;
 
     public BaseConnectionMonitorSubsystem(XPropertyManager propertyManager) {
         log.info("Creating");
@@ -23,18 +23,18 @@ public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem imple
         connectionLatch.addObserver(this);
     }
 
-    public synchronized void setLastPacketReceivedTimeStamp(double currentTimeStamp) {
-        connectionLatch.setValue((currentTimeStamp - lastPacketReceivedTimeStamp) < timeOut.get());
-        this.lastPacketReceivedTimeStamp = currentTimeStamp;
+    public synchronized void setLastPacketReceivedTimestamp(double currentTimestamp) {
+        connectionLatch.setValue((currentTimestamp - lastPacketReceivedTimestamp) < timeOut.get());
+        this.lastPacketReceivedTimestamp = currentTimestamp;
     }
 
     public double getPreviousDisconnectionTime() {
-        return previousDisconnectionTimeStamp;
+        return previousDisconnectionTimestamp;
     }
 
     @Override
     public void update(Observable o, Object arg) {
         log.warn("The Driver Station has been disconnected for greater than " + timeOut + " Second(s)");
-        previousDisconnectionTimeStamp = lastPacketReceivedTimeStamp;
+        previousDisconnectionTimestamp = lastPacketReceivedTimestamp;
     }
 }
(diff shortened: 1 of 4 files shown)


tests/xbot/common/subsystems/ConnectionMonitorTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class ConnectionMonitorTest extends BaseWPITest {
+
+    private MockTimer mockTimer;
+
+    @Override
+    public void setUp() {
+        super.setUp();
+        mockTimer = injector.getInstance(MockTimer.class);
+        mockTimer.advanceTimeInSecondsBy(488);
```
JohnGilb: =]

src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,40 @@
...
+
+import edu.wpi.first.wpilibj.Timer;
+import xbot.common.command.BaseSubsystem;
+import xbot.common.logic.Latch;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+import java.util.Observable;
+import java.util.Observer;
+
```
JohnGilb: This should probably be a singleton

src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,40 @@
...
+public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem implements Observer {
```
JohnGilb: class probably doesn't need to be abstract if you don't have any abstract methods

tests/xbot/common/subsystems/ConnectionMonitorTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void testDisconnectedDriverStation() {
+        BaseConnectionMonitorSubsystem connectionMonitor = injector.getInstance(MockConnectionMonitor.class);
+        mockTimer.advanceTimeInSecondsBy(488);
+        connectionMonitor.setLastPacketReceivedTimestamp(Timer.getFPGATimestamp());
+        assertTrue(connectionMonitor.getPreviousDisconnectionTime() > 400);
+    }
+}
+
```
JohnGilb: I'd like to see some tests that push your system a little more, such as

- small time interval (validate no disconnection message)
- small time interval (validate no disconnection message)
- large time interval (validate disconnection message, last reporting time)
- small time interval (validate no disconnection message, previous reporting time)

src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
```diff
@@ -9,17 +10,18 @@
 import java.util.Observable;
 import java.util.Observer;
 
-public abstract class BaseConnectionMonitorSubsystem extends BaseSubsystem implements Observer {
+@Singleton
+public class BaseConnectionMonitorSubsystem extends BaseSubsystem implements Observer {
```
JohnGilb: Since this isn't abstract, you probably should remove "Base."

If you imagine this being used just for the driver station, I'd call it "DriverStationConnectionMonitorSubsystem" or something like that.

tests/xbot/common/subsystems/ConnectionMonitorTest.java
```diff
@@ -23,18 +22,35 @@ public void setUp() {
     }
```
JohnGilb: These are good tests that you've added, but what I meant was to have at least one test where setLastPacketReceivedTimestamp is called more than once.

Your current tests do a good job of evaluating if your class responds well to some immediate impulse, but they don't test how your class behaves if its primary method is called several times in a row.

src/xbot/common/subsystems/BaseConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,42 @@
...
+    private double lastPacketReceivedTimestamp = Timer.getFPGATimestamp();
+    private double previousDisconnectionTimestamp = -1;
+
+    public BaseConnectionMonitorSubsystem(XPropertyManager propertyManager) {
+        log.info("Creating");
+        timeOut = propertyManager.createPersistentProperty("ConnectionMonitorTimeOut Seconds", 1.0);
+        connectionLatch.addObserver(this);
+    }
+
+    public synchronized void setLastPacketReceivedTimestamp(double currentTimestamp) {
```
WasabiFan: When is this intended to be called, and by whom?

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,46 @@
...
+import xbot.common.properties.XPropertyManager;
+
+import java.util.Observable;
+import java.util.Observer;
+
+@Singleton
+public class ConnectionMonitorSubsystem extends BaseSubsystem implements Observer {
+
+    protected final DoubleProperty timeOut;
+    private final Latch connectionLatch = new Latch(true, Latch.EdgeType.Both);
```
imathno: @JohnGilb @WasabiFan when I choose the Falling edge type here it will only work once when I try to reset the state by forcing a rising edge on the test. It won't detect the next falling edge.

JohnGilb: Does the Latch test class test for that? If not, you should add a test. If it fails, you can fix the Latch. If it passes, then you need to figure out what you're doing differently.

imathno: Fixed

WasabiFan:
Can you merge from master so it's using the latch updates?

imathno:
@WasabiFan I have it merged on my local branch working on it for a bit.

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,42 @@
...
+    private double lastPacketReceivedTimestamp = Timer.getFPGATimestamp();
+    private double previousDisconnectionTimestamp = -1;
+
+    public ConnectionMonitorSubsystem(XPropertyManager propertyManager) {
+        log.info("Creating");
+        timeOut = propertyManager.createPersistentProperty("ConnectionMonitorTimeOut Seconds", 1.0);
+        connectionLatch.addObserver(this);
+    }
+
+    public synchronized void setLastPacketReceivedTimestamp(double currentTimestamp) {
```
WasabiFan: Why `synchronized`? I don't think there are other threads involved here...

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Observable;
+import java.util.Observer;
+
+@Singleton
+public class ConnectionMonitorSubsystem extends BaseSubsystem implements Observer {
+
+    protected final DoubleProperty timeOut;
+    private final Latch connectionLatch = new Latch(true, Latch.EdgeType.FallingEdge);
+
+    private double lastPacketReceivedTimestamp = Timer.getFPGATimestamp();
```
WasabiFan: This will mean that the monitor assumes it was connected whenever the system boots, which seems like the wrong assumption. I'd prefer setting this to `0` or making the field nullable and using `null`.

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,42 @@
...
+        this.lastPacketReceivedTimestamp = currentTimestamp;
+    }
+
+    public synchronized double getPreviousDisconnectionTimestamp() {
+        return previousDisconnectionTimestamp;
+    }
+
+    @Override
+    public void update(Observable o, Object arg) {
+        log.warn("The Driver Station has been disconnected for greater than " + timeOut.get() + " Second(s)");
```
WasabiFan: lowercase `second(s)`

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
WasabiFan: Who is intended to call this method?

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,42 @@
...
+        timeOut = propertyManager.createPersistentProperty("ConnectionMonitorTimeOut Seconds", 1.0);
+        connectionLatch.addObserver(this);
+    }
+
+    public synchronized void setLastPacketReceivedTimestamp(double currentTimestamp) {
+        connectionLatch.setValue((currentTimestamp - lastPacketReceivedTimestamp) < timeOut.get());
+        this.lastPacketReceivedTimestamp = currentTimestamp;
+    }
+
+    public synchronized double getPreviousDisconnectionTimestamp() {
```
WasabiFan: What's the use case for this?

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
imathno: Mainly for testing

imathno: BaseRobot in sharedPeriodic()

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
imathno: For testing

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
WasabiFan: OK... broader question. In total, how is this class intended to be used within a robot? What components call which methods?

imathno: Original idea was anyone can call it but since only BaseRobot is calling it yeah it doesn't make sense for synchronized.

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
imathno: Made it -1

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
imathno: okie

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,43 @@
...
+    private double previousDisconnectionTimestamp = -1;
+
+    public ConnectionMonitorSubsystem(XPropertyManager propertyManager) {
+        log.info("Creating");
+        timeOut = propertyManager.createPersistentProperty("ConnectionMonitor TimeOut Threshold Seconds", 1.0);
+        connectionLatch.addObserver(this);
+    }
+
+    public void setLastPacketReceivedTimestamp(double currentTimestamp) {
+        if (lastPacketReceivedTimestamp != -1) {
```
WasabiFan: 😨 Agh! Never compare floating-point values for equality! Use `< 0` instead here.

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,43 @@
...
+            connectionLatch.setValue((currentTimestamp - lastPacketReceivedTimestamp) < timeOut.get());
+        }
+        this.lastPacketReceivedTimestamp = currentTimestamp;
+    }
+
+    public double getPreviousDisconnectionTimestamp() {
```
WasabiFan: I don't like that this method returns `-1`... it could be confusing/unexpected to a caller. I think storing the state as `null` instead might be better. @JohnGilb Thoughts?

imathno: I think null would be better here.

JohnGilb: I'm torn - I also agree that null is better in most programming, but we often program very "defensively" on the robot, trying to minimize the damage any one mistake can make, and misusing potentially-null objects is a very easy way to cause an exception.

How about sidestepping the problem entirely? Have your public method return "the number of seconds we have been disconnected." This would be 0 most of the time, and a positive double for disconnection scenarios.

In order to preserve the useful timestamp information, you could have this class be responsible for logging the exact time of disconnection.

JohnGilb: Additionally, although you only use this method for testing so far, I can imagine how other parts of the robot might eventually have interesting behavior in disconnection scenarios.

That being said, we still have the fundamental problem of "commands/subsystems don't run if the robot is disconnected" so this is mostly just aggressive future-proofing, and using null is probably fine for now.

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
```diff
@@ -0,0 +1,43 @@
...
+        return previousDisconnectionTimestamp;
+    }
+
+    @Override
+    public void update(Observable o, Object arg) {
+        log.warn("The Driver Station has been disconnected for greater than " + timeOut.get() + " second(s)");
```
JohnGilb: Given your current implementation (since SharedPeriodic is only called when we get a data packet from the Driver Station), it might be better to say here:

"Connection with Driver Station restored. The Robot was disconnected from the driver station for at least" + timeOut.get() + " second(s)."

src/xbot/common/subsystems/ConnectionMonitorSubsystem.java
WasabiFan: Yeah... What is the expected behavior of this code? Is it just for logging? I'm not sure any of this code will run anyway, unless we out it on its own timer on a different thread... So even logging won't ever be seen/used.

imathno: Maybe we can add a notify() method in the base subsystem and the subsystem can override the method to do what they desire with the information. Subsystems maybe can do ConnectionMonitorSubsystem.register(Subsystem);

imathno: We can implement more logic to detect trends and use all the other systems to see why we're disconnecting and try to fix itself.

WasabiFan: What I'd like to know is: what is the action that is taken in response to the disconnection? And this code requires a Boolean input indicating whether the robot is connected; where does that Boolean come from?

Separately, we must consider whether this code will ever actually run. The robot loop runs only while it is connected to the driver station; so as far as I can tell, it's just dead code.

imathno: Lets leave this PR open and see what we can use this information for.

imathno: But for now I see what you're saying and the code will never get executed. Only in a simulated environment like the jUnit test it looks viable.

JohnGilb: In theory this could be used by the robot on a reconnection - it'll notice that it didn't get any messages for a long time and just got reconnected, and at the very least we'll have a robot-side log of that particular event.

Perhaps other subsystems might want to take special action if they knew they had been suspended for a long time?

WasabiFan: I think both of those are interesting scenarios; the logic here would need to change though. It would be tracking time since last check rather than latching on an `isConnected` flag, and we'd probably want a notification system line @imathno mentioned where it alerted registered subsystems when operation resumed.

Alternately, we could add a logging latch to the loop frequency counter in [BaseRobot.java](https://github.com/Team488/SeriouslyCommonLib/blob/master/src/xbot/common/command/BaseRobot.java) so it logs whenever loop frequency drops below a threshold. I'm liking this latter idea the more I think about it.

JohnGilb:

PR #106, opened 2017-12-26
Fix Latch logic

imathno:
Added Latch Tests

We didn't save the previous value so saveValueAndAlertWatchers was only being called once.

### DIFF
--- a/src/xbot/common/logic/Latch.java
+++ b/src/xbot/common/logic/Latch.java
@@ -7,8 +7,8 @@ import java.util.Observable;
  */
 public class Latch extends Observable {
 
+    private final EdgeType latchType;
     private boolean value;
-    private EdgeType latchType;
 
     public enum EdgeType {
         RisingEdge, FallingEdge, Both
@@ -20,31 +20,13 @@ public class Latch extends Observable {
     }
 
     public void setValue(boolean newValue) {
-        if (this.value == newValue) {
-            return; /* Do nothing if value stays the same */
-        }
+        if (this.value == newValue) return; /* Do nothing if value stays the same */
 
-        this.value = newValue;
         EdgeType edgeType = newValue ? EdgeType.RisingEdge : EdgeType.FallingEdge;
-        boolean alertWatchers;
-        switch (latchType) {
-            case RisingEdge:
-                alertWatchers = newValue; /* Alert Watchers if newValue equals true */
-                break;
-            case FallingEdge:
-                alertWatchers = !newValue; /* Alert Watchers if newValue equals false */
-                break;
-            case Both:
-                alertWatchers = true;
-                break;
-            default:
-                alertWatchers = false;
-                break;
-        }
-
-        if (alertWatchers) {
+        if (latchType == edgeType || latchType == EdgeType.Both) {
             setChanged();
             notifyObservers(edgeType);
         }
+        this.value = newValue;
     }
 }
--- a/src/xbot/common/logic/Latch.java
+++ b/src/xbot/common/logic/Latch.java
@@ -22,30 +22,28 @@ public class Latch extends Observable {
     public void setValue(boolean value) {
         boolean different = this.value != value;
 
-        switch (latchType) {
-        case Both:
-            if (different) {
-                saveValueAndAlertWatchers(value);
+        if (different) {
+            switch (latchType) {
+                case Both:
+                    saveValueAndAlertWatchers(value);
+                    break;
+                case RisingEdge:
+                    if (value)
+                        saveValueAndAlertWatchers(true);
+                    break;
+                case FallingEdge:
+                    if (!value) {
+                        saveValueAndAlertWatchers(false);
+                    }
+                    break;
+                default:
+                    this.value = value;
+                    break;
             }
-            break;
-        case RisingEdge:
-            if (different && value) {
-                saveValueAndAlertWatchers(value);
-            }
-            break;
-        case FallingEdge:
-            if (different && !value) {
-                saveValueAndAlertWatchers(value);
-            }
-            break;
-        default:
-            break;
         }
     }
 
     private void saveValueAndAlertWatchers(boolean value) {
-        this.value = value;
-
         EdgeType edgeType = value ? EdgeType.RisingEdge : EdgeType.FallingEdge;
 
         setChanged();
--- a/src/xbot/common/logic/Latch.java
+++ b/src/xbot/common/logic/Latch.java
@@ -23,28 +23,28 @@ public class Latch extends Observable {
         boolean different = this.value != value;
 
         if (different) {
+            this.value = value;
             switch (latchType) {
                 case Both:
-                    saveValueAndAlertWatchers(value);
+                    alertWatchers(value);
                     break;
                 case RisingEdge:
                     if (value) {
-                        saveValueAndAlertWatchers(true);
+                        alertWatchers(true);
                     }
                     break;
                 case FallingEdge:
                     if (!value) {
-                        saveValueAndAlertWatchers(false);
+                        alertWatchers(false);
                     }
                     break;
                 default:
-                    this.value = value;
                     break;
             }
         }
     }
 
-    private void saveValueAndAlertWatchers(boolean value) {
+    private void alertWatchers(boolean value) {
         EdgeType edgeType = value ? EdgeType.RisingEdge : EdgeType.FallingEdge;
 
         setChanged();
--- a/src/xbot/common/logic/Latch.java
+++ b/src/xbot/common/logic/Latch.java
@@ -28,8 +28,9 @@ public class Latch extends Observable {
                     saveValueAndAlertWatchers(value);
                     break;
                 case RisingEdge:
-                    if (value)
+                    if (value) {
                         saveValueAndAlertWatchers(true);
+                    }
                     break;
                 case FallingEdge:
                     if (!value) {
--- a/src/xbot/common/logic/Latch.java
+++ b/src/xbot/common/logic/Latch.java
@@ -20,7 +20,9 @@ public class Latch extends Observable {
     }
 
     public void setValue(boolean newValue) {
-        if (this.value == newValue) return; /* Do nothing if value stays the same */
+        if (this.value == newValue) {
+            return; /* Do nothing if value stays the same */
+        }
 
         EdgeType edgeType = newValue ? EdgeType.RisingEdge : EdgeType.FallingEdge;
         if (latchType == edgeType || latchType == EdgeType.Both) {
--- a/src/xbot/common/logic/Latch.java
+++ b/src/xbot/common/logic/Latch.java
@@ -26,7 +26,7 @@ public class Latch extends Observable {
 
         this.value = newValue;
         EdgeType edgeType = newValue ? EdgeType.RisingEdge : EdgeType.FallingEdge;
-        boolean alertWatchers = false;
+        boolean alertWatchers;
         switch (latchType) {
             case RisingEdge:
                 alertWatchers = newValue; /* Alert Watchers if newValue equals true */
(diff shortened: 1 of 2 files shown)


src/xbot/common/logic/Latch.java
```diff
@@ -22,33 +22,32 @@ public Latch(boolean initialValue, EdgeType latchType) {
...
+                        saveValueAndAlertWatchers(true);
+                    }
+                    break;
+                case FallingEdge:
+                    if (!value) {
+                        saveValueAndAlertWatchers(false);
+                    }
+                    break;
+                default:
+                    this.value = value;
```
WasabiFan: Ummm... are you sure you want to put this in `default`? It will never run.

src/xbot/common/logic/Latch.java
```diff
@@ -22,33 +22,32 @@ public Latch(boolean initialValue, EdgeType latchType) {
...
-                saveValueAndAlertWatchers(value);
-            }
-            break;
-        default:
-            break;
         }
     }
 
     private void saveValueAndAlertWatchers(boolean value) {
-        this.value = value;
```
WasabiFan: What am I missing -- isn't this where we saved the previous value?

tests/xbot/common/logic/LatchTest.java
```diff
@@ -0,0 +1,76 @@
...
+        latch.setValue(false);
+        assertTrue(mockLatch.getEdgeType() == Latch.EdgeType.FallingEdge);
+        latch.setValue(true);
+        assertTrue(mockLatch.getEdgeType() == Latch.EdgeType.FallingEdge);
+        latch.setValue(false);
+        assertTrue(mockLatch.getNumTimesUpdated() == 2);
+    }
+}
+
+class MockLatch implements Observer {
```
WasabiFan: I think this class should be renamed `LatchTestObserver` or similar. It would be good if `getEdgeType` were called `getLastUpdateEdgeType` to make it clear that it is reporting the state from the Latch.

tests/xbot/common/logic/LatchTest.java
```diff
@@ -0,0 +1,76 @@
...
+import static junit.framework.TestCase.assertTrue;
+
+public class LatchTest {
+
+    @Test
+    public void testRisingEdge() {
+        Latch latch = new Latch(false, Latch.EdgeType.RisingEdge);
+        MockLatch mockLatch = new MockLatch();
+        latch.addObserver(mockLatch);
+        latch.setValue(true);
```
WasabiFan: Assert before and after calling `setValue`

tests/xbot/common/logic/LatchTest.java
```diff
@@ -0,0 +1,76 @@
...
+        assertTrue(mockLatch.getEdgeType() == Latch.EdgeType.RisingEdge);
```
WasabiFan: `assertEqual`?

src/xbot/common/logic/Latch.java
```diff
@@ -23,28 +23,28 @@ public void setValue(boolean value) {
...
                 default:
-                    this.value = value;
                     break;
             }
         }
     }
 
-    private void saveValueAndAlertWatchers(boolean value) {
+    private void alertWatchers(boolean value) {
         EdgeType edgeType = value ? EdgeType.RisingEdge : EdgeType.FallingEdge;
```
WasabiFan: Can this method just accept an `EdgeType` so we don't need to extrapolate like this?

tests/xbot/common/logic/LatchTest.java
```diff
@@ -14,8 +14,9 @@ public void testRisingEdge() {
         Latch latch = new Latch(false, Latch.EdgeType.RisingEdge);
         MockLatch mockLatch = new MockLatch();
         latch.addObserver(mockLatch);
+        assertEquals(mockLatch.getEdgeType(), null);
         latch.setValue(true);
-        assertTrue(mockLatch.getEdgeType() == Latch.EdgeType.RisingEdge);
+        assertEquals(mockLatch.getEdgeType(), Latch.EdgeType.RisingEdge);
```
WasabiFan: is it `actual, expected` or `expected, actual`? I think these args might be reversed.

src/xbot/common/logic/Latch.java
imathno: oops some how that passed the test

src/xbot/common/logic/Latch.java
JohnGilb: Agreed - you do all the work to figure out the edge type in setValue, but then throw it away and figure it out again in alertWatchers. =]

tests/xbot/common/logic/LatchTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.Test;
+
+import java.util.Observable;
+import java.util.Observer;
+
+import static junit.framework.TestCase.assertEquals;
+
+public class LatchTest {
+
+    @Test
```
JohnGilb: You don't have any tests that set the same value more than once, so you'll never exercise one branch of the "isDifferent" code in latch.

(It might be a good idea to go to Help -> Eclipse Marketplace and install EclEmma, a code coverage tool, and you can see what code never gets executed)

src/xbot/common/logic/Latch.java
```diff
@@ -22,33 +22,32 @@ public Latch(boolean initialValue, EdgeType latchType) {
...
-            break;
-        default:
-            break;
         }
     }
 
-    private void saveValueAndAlertWatchers(boolean value) {
-        this.value = value;
-
+    private void alertWatchers(boolean value) {
```
WasabiFan: Accept edge type as parameter instead of boolean

src/xbot/common/logic/Latch.java
```diff
@@ -22,33 +22,32 @@ public Latch(boolean initialValue, EdgeType latchType) {
     public void setValue(boolean value) {
```
WasabiFan: call this parameter `newValue`?

tests/xbot/common/logic/LatchTest.java
```diff
@@ -0,0 +1,133 @@
...
+    public void testRisingEdge() {
+        Latch latch = new Latch(false, Latch.EdgeType.RisingEdge);
+        LatchTestObserver latchTestObserver = new LatchTestObserver();
+        latch.addObserver(latchTestObserver);
+        assertEquals(latchTestObserver.getLastUpdateEdgeType(), null);
+        latch.setValue(true);
+        assertEquals(latchTestObserver.getLastUpdateEdgeType(), Latch.EdgeType.RisingEdge);
```
WasabiFan: Aren't these method parameters supposed to be `expected, actual`? I can't remember exactly.

tests/xbot/common/logic/LatchTest.java
```diff
@@ -0,0 +1,133 @@
...
+        assertEquals(latchTestObserver.getNumTimesUpdated(), 2);
+    }
+
+    @Test
+    public void testFallRiseFallObservingFallingEdge() {
+        Latch latch = new Latch(true, Latch.EdgeType.FallingEdge);
+        LatchTestObserver latchTestObserver = new LatchTestObserver();
+        latch.addObserver(latchTestObserver);
+        latch.setValue(false);
+        assertEquals(latchTestObserver.getLastUpdateEdgeType(), Latch.EdgeType.FallingEdge);
```
WasabiFan: Can we assert on the number of updates at each step (all tests)? Part of the contract of the latch is that it only fires an update when the edge actually happens; I'm slightly worried that two complementary bugs could cancel each other out or a logic inversion could be introduced and the tests would still pass.

tests/xbot/common/logic/LatchTest.java
```diff
@@ -0,0 +1,133 @@
...
+        return lastUpdateEdgeType;
+    }
+
+    public int getNumTimesUpdated() {
+        return numTimesUpdated;
+    }
+
+    @Override
+    public void update(Observable o, Object arg) {
+        numTimesUpdated++;
```
WasabiFan: Can we assert here that `arg` is an `EdgeType`? It'll throw a `ClassCastException` anyway, but it's easier to track down an assertion failure than a CCE.

JohnGilb:
Good policy is to reply to each comment on your pull request with at least some minimum information like:
- "Fixed this"
- "Not going to fix this" (and some explanation of why)

That way we know which comments you believe you have addressed with your changes, which lets reviewers go over your PR much faster, and avoid having issues accidentally get ignored by both parties.

tests/xbot/common/logic/LatchTest.java
imathno: So do (arg instanceof EdgeType) here?

WasabiFan: Yeah, and I think you can include a message too:

```java
assertTrue(arg instanceof EdgeType, "Latch must provide an argument of type EdgeType");
```

src/xbot/common/logic/Latch.java
imathno: okie

src/xbot/common/logic/Latch.java
imathno: I just removed the alertWatchers method and handle everything in setValue

tests/xbot/common/logic/LatchTest.java
imathno: I don't write much test so I forgot

tests/xbot/common/logic/LatchTest.java
imathno: Added

tests/xbot/common/logic/LatchTest.java
imathno: Added

src/xbot/common/logic/Latch.java
```diff
@@ -19,36 +19,32 @@ public Latch(boolean initialValue, EdgeType latchType) {
...
-        boolean different = this.value != value;
+    public void setValue(boolean newValue) {
+        if (this.value == newValue) {
+            return; /* Do nothing if value stays the same */
+        }
 
+        this.value = newValue;
+        EdgeType edgeType = newValue ? EdgeType.RisingEdge : EdgeType.FallingEdge;
+        boolean alertWatchers;
         switch (latchType) {
```
WasabiFan: Instead of this switch statement, can't we just alert if `latchType.equals(LatchType.Both) || latchType.equals(edgeType)`? Not 100% sure I logicked good as it's currently 4:00 in the morning.

src/xbot/common/logic/Latch.java
```diff
@@ -19,36 +19,32 @@ public Latch(boolean initialValue, EdgeType latchType) {
         this.latchType = latchType;
     }
 
-    public void setValue(boolean value) {
-        boolean different = this.value != value;
+    public void setValue(boolean newValue) {
+        if (this.value == newValue) {
```
WasabiFan: Nice! :+1:

src/xbot/common/logic/Latch.java
imathno: Some hard thinking I figured out how to shorten most things.

src/xbot/common/logic/Latch.java
```diff
@@ -20,31 +20,13 @@ public Latch(boolean initialValue, EdgeType latchType) {
...
-            case Both:
-                alertWatchers = true;
-                break;
-            default:
-                alertWatchers = false;
-                break;
-        }
-
-        if (alertWatchers) {
+        if (latchType == edgeType || latchType == EdgeType.Both) {
```
WasabiFan: Can you do that? I'm not sure if enum equality works like that in Java...

WasabiFan: Never mind, yeah that looks right: https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals

WasabiFan:
Looks great! Much cleaner than it was before.

PR #112, opened 2018-01-14
Feature/add gamepad

JohnGilb:


src/main/java/xbot/common/controls/sensors/XGamepad.java
```diff
@@ -0,0 +1,70 @@
...
+import xbot.common.injection.wpi_factories.CommonLibFactory;
+import xbot.common.logging.RobotAssertionManager;
+import xbot.common.math.XYPair;
+
+public abstract class XGamepad extends XJoystick {
+
+	// All of the "Left Joystick" inversions are kept track of by the base XJoystick class.
+	// We just need to redirect any calls to the base class.
+	private boolean xRightInverted = false;
+    private boolean yRightInverted = false;
```
WasabiFan: Indentation

src/main/java/xbot/common/controls/sensors/XGamepad.java
```diff
@@ -0,0 +1,70 @@
+package xbot.common.controls.sensors;
+
+import xbot.common.injection.wpi_factories.CommonLibFactory;
+import xbot.common.logging.RobotAssertionManager;
+import xbot.common.math.XYPair;
+
+public abstract class XGamepad extends XJoystick {
+
+	// All of the "Left Joystick" inversions are kept track of by the base XJoystick class.
```
WasabiFan: 😒

src/main/java/xbot/common/controls/sensors/XGamepad.java
```diff
@@ -0,0 +1,70 @@
...
+	// We just need to redirect any calls to the base class.
+	private boolean xRightInverted = false;
+    private boolean yRightInverted = false;
+
+	public XGamepad(int port, CommonLibFactory clf, RobotAssertionManager assertionManager, int numButtons) {
+		super(port, clf, assertionManager, numButtons);
+	}
+	
+	public XYPair getRightVector() {
```
WasabiFan: indentation

src/main/java/xbot/common/controls/sensors/XGamepad.java
```diff
@@ -0,0 +1,70 @@
...
+        return new XYPair(getRightStickX(), getRightStickY());
+    }
+
+    public XYPair getLeftVector() {
+        return this.getVector();
+    }
+    
+    public double getRightStickX(){
+        return this.getRawAxis(4) * (getRightStickXInversion() ? -1 : 1);
```
WasabiFan: My recollection is that these axes are inconsistent between Windows versions/drivers; where did these numbers come from?

src/main/java/xbot/common/controls/sensors/mock_adapters/MockGamepad.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.inject.assistedinject.Assisted;
+
+import xbot.common.controls.sensors.XGamepad;
+import xbot.common.injection.wpi_factories.CommonLibFactory;
+import xbot.common.logging.RobotAssertionManager;
+
+public class MockGamepad extends XGamepad {
+
+	Map<Integer, Boolean> buttons = new HashMap<Integer, Boolean>();
+    Map<Integer, Double> rawAxis = new HashMap<Integer, Double>();
```
WasabiFan: Indentation

src/main/java/xbot/common/controls/sensors/wpi_adapters/GamepadWpiAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+
+import edu.wpi.first.wpilibj.GenericHID;
+import edu.wpi.first.wpilibj.Joystick;
+import xbot.common.controls.sensors.XGamepad;
+import xbot.common.injection.wpi_factories.CommonLibFactory;
+import xbot.common.logging.RobotAssertionManager;
+
+public class GamepadWpiAdapter extends XGamepad {
+
+private GenericHID internalHID;
```
WasabiFan: Indentation

src/main/java/xbot/common/controls/sensors/wpi_adapters/GamepadWpiAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+            @Assisted("port") int port, 
+            @Assisted("numButtons") int numButtons,
+            CommonLibFactory clf,
+            RobotAssertionManager assertionManager) {
+        super(port, clf, assertionManager, numButtons);
+        
+        internalHID = new Joystick(port);
+    }
+    
+    protected double getX() {
```
WasabiFan: Why `protected`?

src/main/java/xbot/common/injection/wpi_factories/CommonLibFactory.java
```diff
@@ -35,6 +36,10 @@ public XJoystick createJoystick(
             @Assisted("port") int port,
             @Assisted("numButtons") int numButtons);
     
+    public XGamepad createGamepad(
+    		@Assisted("port") int port,
```
WasabiFan: Indentation

JohnGilb:
@WasabiFan

> The fact that one stick is handled by WPILib's logic and the other by ours feels wrong. but at the same time I feel like we have to support it because it is a joystick implementation.

That's not exactly true - what we're extending is XJoystick, not the WPI joystick. What this is saying is that "you can treat this just like a classic joystick by calling getVector(), but that will only return the left stick. If you want to treat this like a gamepad, call getLeftVector and getRightVector". 

Since all the existing XJoystick code already handles inversion and such for the left joystick, we just reuse those calls.

src/main/java/xbot/common/controls/sensors/XGamepad.java
JohnGilb: I think something is wrong with whitespace on one of my machines. I just did a Source->Correct Indentation, we'll see how that's interpreted by github.

src/main/java/xbot/common/controls/sensors/XGamepad.java
JohnGilb: I should actually be more specific - this is a FTCGamepad implementation (or whatever the proper brand name of the FTC gamepads is). Those have been very consistent. I believe it's the Xbox controllers that have been inconsistent in the past.

src/main/java/xbot/common/controls/sensors/wpi_adapters/GamepadWpiAdapter.java
JohnGilb: See XJoystick - the primitive getX and getY are implemented by the adapters, and the abstract class uses them to produce things like vectors.

PR #115, opened 2018-01-15
Adds a velocity-decay version of the HeadingAssistModule

JohnGilb:
I've been doing some research into how other teams add automation to the drivetrain, and in a couple places I saw a pattern of using a gyro to automatically add "turning resistance" to the robot, so that when the driver stops rotating, the robot stops turning much faster.

The basic mechanism is to use a D-only PID on the gyro.

I've added this as a variant into our HeadingAssistModule - now we can choose between "hold an orientation after rotating" and "rapidly slow down after rotating"

### DIFF
--- a/src/main/java/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
+++ b/src/main/java/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
@@ -17,6 +17,10 @@ import xbot.common.subsystems.pose.BasePoseSubsystem;
  * - If the human driver wants the robot to turn (above some small threshold), then the robot should turn
  *   according to human desires.
  * - After a few moments, the robot should memorize its current heading, and attempt to hold it again.
+ * 
+ * The automatic response can take one of two forms:
+ * - Hold a specific heading, as mentioned above, using a PID controller
+ * - Resist changes in rotational velocity using a D-only controller.
  * @author jogilber
  *
  */
--- a/src/main/java/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
+++ b/src/main/java/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
@@ -22,22 +22,36 @@ public class HeadingAssistModule {
 
     final BasePoseSubsystem pose;
     final HeadingModule headingModule;
+    final HeadingModule decayModule;
+    
     final DoubleProperty humanThreshold;
     final DoubleProperty coastTime;
     double desiredHeading;
     double lastHumanInput;
     boolean inAutomaticMode;
+    private HeadingAssistMode headingMode;
+    
+    public enum HeadingAssistMode {
+        HoldOrientation,
+        DecayVelocity
+    }
     
     @AssistedInject
     public HeadingAssistModule(
             @Assisted("headingModule") HeadingModule headingModule,
+            @Assisted("decayModule") HeadingModule decayModule,
             XPropertyManager propMan,
             BasePoseSubsystem pose) {
         this.headingModule = headingModule;
+        this.decayModule = decayModule;
         this.pose = pose;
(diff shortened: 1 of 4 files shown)


src/main/java/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
```diff
@@ -22,24 +22,38 @@
...
         this.headingModule = headingModule;
+        this.decayModule = decayModule;
         this.pose = pose;
         humanThreshold = propMan.createPersistentProperty("HeadingAssistModule - Human Threshold", 0.05);
         coastTime = propMan.createPersistentProperty("Heading Assist Module - Coast Time", 0.5);
         lastHumanInput = 0;
     }
     
+    public void setMode(HeadingAssistMode mode) {
+        this.headingMode = mode;
```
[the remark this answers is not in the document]
JohnGilb: There's a switch statement further down that checks the mode.

However, you did make me realize there was no protection against null objects, so I've fixed that.

src/main/java/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
```diff
@@ -72,11 +86,19 @@ else if (timeSinceHumanInput >= coastTime.get() && !inAutomaticMode) {
...
             inAutomaticMode = true;
             headingModule.reset();
+            decayModule.reset();
             return 0;
         }
         // by this point, the only option left is that the machine is in automatic mode.
         else {
-            return headingModule.calculateHeadingPower(desiredHeading);
+            switch (headingMode) {
+            case HoldOrientation:
```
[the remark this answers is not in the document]
JohnGilb: Fixed, I believe. I'm getting suspicious of eclipse auto-indent. =]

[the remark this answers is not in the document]
WasabiFan: _this particular instance_ is a painfully common intentional formatting style of switch statements; I change the formatter settings to fix this

src/main/java/xbot/common/subsystems/drive/control_logic/HeadingAssistModule.java
```diff
@@ -52,31 +71,44 @@ public HeadingAssistModule(
...
+            if (headingMode == null) {
+                log.warn("Heading module did not have any mode set. Forcing to HoldOrientation.");
+                this.headingMode = HeadingAssistMode.HoldOrientation;
+            }
+
+            switch (headingMode) {
+                case HoldOrientation:
+                    return headingModule.calculateHeadingPower(desiredHeading);
+                case DecayVelocity:
+                    return decayModule.calculateHeadingPower(0);
```
WasabiFan: I still don't understand how this works. This function takes a desired absolute heading. If you call it with 0, won't it just rotate to heading 0? And in the OP, you said it specifically was designed to use only D in the controller. How is that done?

JohnGilb: the DecayModule's PID parameters are set outside of this class - for an example, see HeadingAssistModuleTest:

```
HeadingModule hold = clf.createHeadingModule(pf.createPIDManager("Hold", 1000, 0, 0));
        HeadingModule decay = clf.createHeadingModule(pf.createPIDManager("Decay", 0, 0, 1000));
        clf.createHeadingAssistModule(hold, decay);
```

A PID controller with only D will just resist changes in error, regardless of the target.

WasabiFan: Ah, I get it; I completely missed that those two parameters are assisted rather than injected. Can you add some Javadoc on the `HeadingModule` and `HeadingAssistModule` to explain what they do and how they're different?

JohnGilb: Great idea! Will do.

PR #120, opened 2018-01-19
MockCANTalons were missing some implementation for feedback sensors.

JohnGilb:


src/main/java/xbot/common/controls/actuators/mock_adapters/MockCANTalon.java
```diff
@@ -278,7 +278,7 @@ public ErrorCode configSelectedFeedbackSensor(FeedbackDevice feedbackDevice, int
     @Override
     public int getSelectedSensorPosition(int pidIdx) {
         // TODO Auto-generated method stub
```
WasabiFan: .

PR #121, opened 2018-01-20
Add ability to run command while button is pressed, but not re-start it after it finishes

WasabiFan:


src/main/java/xbot/common/logic/Latch.java
```diff
@@ -19,6 +22,15 @@ public Latch(boolean initialValue, EdgeType latchType) {
         this.latchType = latchType;
     }
 
+    public Latch(boolean initialValue, EdgeType latchType, Consumer<EdgeType> callback) {
```
WasabiFan: The latch changes are to simplify usage. Whereas before you had to supply an `Observer`, which was passed a generally useless "sender" and an `Object` you had to cast for use, now you can pass your callback to the constructor (or add it later as before) without having to do the cast.

JohnGilb: This is a good improvement!

src/main/java/xbot/common/controls/sensors/AdvancedButton.java
```diff
@@ -1,19 +1,45 @@
...
     }
-    
+
     public boolean getInverted() {
         return isInverted;
     }
+
+    public void whilePressedNoRestart(final Command command) {
+        // Based on source from WPILib's Trigger
+        new ButtonScheduler() {
```
JohnGilb: Great to see Latch being used!

PR #123, opened 2018-01-21
Add checkstyle to Gradle build and run it on Travis

WasabiFan:
In doing this I upgraded our checkstyle config file to handle breaking changes in checkstyle 8.0+. Also, I removed the check for sequences of capital letters in class names and added a check for "auto-generated" TODOs.

### DIFF
--- a/src/main/java/edu/wpi/first/wpilibj/MockLidarLite.java
+++ b/src/main/java/edu/wpi/first/wpilibj/MockLidarLite.java
@@ -12,7 +12,6 @@ public class MockLidarLite extends XLidarLite {
     @Inject
     public MockLidarLite(@Assisted("port") Port port, XPropertyManager propMan) {
         super(port, propMan);
-        // TODO Auto-generated constructor stub
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


src/main/java/edu/wpi/first/wpilibj/MockLidarLite.java
```diff
@@ -21,7 +20,6 @@ protected void update() {
         try {
             Thread.sleep(5);
         } catch (InterruptedException e) {
-            // TODO Auto-generated catch block
```
WasabiFan: Is this what we want to do?

src/main/java/edu/wpi/first/wpilibj/MockXboxControllerAdapter.java
```diff
@@ -32,19 +32,16 @@ public MockXboxControllerAdapter(@Assisted("port") int port, RobotAssertionManag
 
     @Override
     protected double getTriggerAxis(Hand hand) {
-        // TODO Auto-generated method stub
```
WasabiFan: Should these have implementations?

JohnGilb: Created a work item to track this - we'll need this if we end up doing full end-to-end tests from button pressing -> robot action.

PR #124, opened 2018-01-21
Fully support our usage of Talons in VictorAppearingAsTalonWpiAdapter

WasabiFan:
At the moment, it'll have lots of funky behavior, such as followers running at full power all the time.

This hasn't been tested on-robot.

### DIFF
--- a/src/main/java/xbot/common/controls/actuators/wpi_adapters/VictorAppearingAsTalonWPIAdapter.java
+++ b/src/main/java/xbot/common/controls/actuators/wpi_adapters/VictorAppearingAsTalonWPIAdapter.java
@@ -33,9 +33,6 @@ public class VictorAppearingAsTalonWPIAdapter extends XCANTalon {
     RobotAssertionManager assertionManager;
     SpeedController internalSpeedController;
     int deviceId;
-    
-    int masterId = -1;
-    SpeedController masterController = null;
 
     @Inject
     public VictorAppearingAsTalonWPIAdapter(@Assisted("deviceId") int deviceId, RobotAssertionManager assertionManager, XPropertyManager propMan) {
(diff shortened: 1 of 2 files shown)


src/main/java/xbot/common/controls/actuators/wpi_adapters/VictorAppearingAsTalonWPIAdapter.java
```diff
@@ -26,29 +26,58 @@
...
         super(deviceId, propMan);
         internalSpeedController = new Victor(deviceId);
         this.deviceId = deviceId;
+        this.assertionManager = assertionManager; 
     }
 
     @Override
     public void set(ControlMode Mode, double demand) {
-        internalSpeedController.set(demand);
+        if(Mode == ControlMode.Follower) {
```
JohnGilb: I think we may just have to avoid using Follow mode entirely if we're using Victors, since Follow is never actually called in set() during normal behavior.

As an example in the DriveSubsystem:

```
        this.leftMaster = factory.createCANTalon(20);
        this.leftFollower = factory.createCANTalon(21);
        leftFollower.follow(leftMaster);
```

And after that, nobody ever calls `.set()` on `leftFollower`.

JohnGilb: Maybe we could spin off a new thread that tries to constantly match speed, but I think for the very short term we can just promote follower talons to master talons when using the Robox.

WasabiFan: Hmmm, good point; this is backwards. We really should be doing what is done in the mock implementation (IIRC) where the mock IO class tracks all Talon instances and allows the master to set the slave. If you'd prefer to promote the followers to masters I'm fine with that, however we should include the assertion I added here either way to enforce that. I had originally intended to _just_ add that assertion in this PR before I realized the follower situation would be a problem.

JohnGilb: Yes, the assertion alone is a good idea.

PR #125, opened 2018-01-24
Allow for building sources with commonlib,

JohnGilb:
and make the MockFTCGamepad more useable by tests


WasabiFan:
Now that I think of it, why _are_ we using the PS2 gamepads rather than the newer Xbox ones we have?

JohnGilb:
Just a matter of current convenience - I imagine we will probably swap to Xbox controllers when the Drive team has a choice.

I also wanted to get a sense of their reliability - the Xbox controllers are clearly better in nearly every way, but we ran into a few connection issues last year that caused some anxiety on the field. These could make a good backup.

PR #127, opened 2018-01-26
Feature/prevent multiple allocation

JohnGilb:
Introduces a new Singleton, the DevicePolice. Devices register there on creation, and it will throw exceptions in tests if it notices double-allocation.


src/main/java/xbot/common/injection/wpi_factories/DevicePolice.java
```diff
@@ -0,0 +1,43 @@
...
+        Analog,
+        SPI,
+        I2C,
+        USB
+    }
+    
+    @Inject
+    public DevicePolice(RobotAssertionManager assertionManager) {
+        this.assertionManager = assertionManager;
+        registeredDevices = new ArrayList<String>();
```
WasabiFan: `LinkedList`?

src/test/java/xbot/common/injection/wpi_factories/TestCommonLibFactory.java
```diff
@@ -19,18 +21,30 @@ public void makeOneOfEverything() {
...
+        clf.createGamepad(3, 10);
         
         clf.createHeadingModule(pf.createPIDManager("bar", 1, 0, 0));
     }
+    
+    @Test
+    public void doubleAllocate() {
+        clf.createCANTalon(1);
+        
+        try {
```
WasabiFan: Rather than catch the exception, you can add an annotation which tells JUnit to expect an exception and fail if it isn't thrown.

JohnGilb: Ooh, nice. I'll look that up.

src/main/java/xbot/common/injection/wpi_factories/DevicePolice.java
JohnGilb: Good idea.

PR #134, opened 2018-02-09
Digital input should support inversion

JohnGilb:


src/main/java/edu/wpi/first/wpilibj/MockDigitalInput.java
```diff
@@ -18,10 +18,16 @@ public MockDigitalInput(@Assisted("channel") int channel, DevicePolice police) {
     }
 
     public void set_value(boolean value) {
```
WasabiFan: While we're here, can we fix this name?

PR #137, opened 2018-02-16
Create a timelogger to help us investigate robot performance

JohnGilb:


src/main/java/xbot/common/command/BaseCommand.java
```diff
@@ -12,12 +13,14 @@
...
 
     protected Logger log;
+    protected TimeLogger monitor;
     
     @Inject
     SmartDashboardCommandPutter commandPutter;
 
     public BaseCommand() {
         log = Logger.getLogger(this.getName());
+        monitor = new TimeLogger(this.getName(), 20);
```
WasabiFan: Um... Did you intend to call into this somewhere?

JohnGilb: Just making it available to child classes, like the logger. Sadly, the way things are designed, I can't override any interesting methods like `run ()` on Command since it's a private method.

PR #143, opened 2018-02-18
square and retain sign

JohnGilb:


src/main/java/xbot/common/math/MathUtils.java
```diff
@@ -24,4 +24,9 @@ public static double scaleDouble(double value, double oldMin, double oldMax, dou
     public static double constrainDoubleToRobotScale(double value) {
         return constrainDouble(value, -1, 1);
     }
+    
+    public static double squareAndRetainSign(double value) {
```
WasabiFan: Could we make this generic for any power?

PR #153, opened 2018-02-28
Feature/add auto delay command

JohnGilb:


WasabiFan:
Could you explain this a bit? Where is the delay command intended to be used.?

JohnGilb:
Mostly autonomous programs - so we can have a delay controlled by some outside source (like a property, or a subsystem) before starting our auto proper.

PR #154, opened 2018-03-01
Restructure PurePursuitCommand to allow custom implementations

WasabiFan:
Existing usages of `PurePursuitCommand` will now instead use `ConfigurablePurePursuitCommand` (will open a PR in 2018 repo). Now the `PurePursuitCommand` is an abstract class which we can subclass for individual uses.


JohnGilb:
Is this meant to lay the groundwork for variations that do things like shift gears, or something else?

WasabiFan:
My goal is to be able to have command classes which produce some points, perhaps perform initialization logic, then drive with PurePursuit. In my case I could implement a "PurePursuitToVisibleCube" command, for example, which internally has the logic to produce a goal pose from a relative detected cube point. It just moves the supplier to a class implementation rather than external configuration.

PR #157, opened 2018-03-03
Allow accessing internal HID in joystick objects

WasabiFan:
This is intended to allow rumbling of devices we use as plain joysticks but know in reality are rumble-able controllers. This is pretty much a hack. For the alternative (which I want to look at _after_ competition as it makes some code cleaner), see https://github.com/Team488/SeriouslyCommonLib/tree/xbox-controller-rumble-dpad and https://github.com/Team488/TeamXbot2018/tree/adopt-xbox-controller-class. The primary issue here is that the Xbox controller class's inheritance tree is entirely separate from our other HIDs.


WasabiFan:
I confirmed from looking at the WPILib implementation that it does the same thing internally whether it's an Xbox controller or not. In fact, the rumble code is in the base `Joystick` class.

PR #161, opened 2018-03-11
Fixed a bug in PurePursuitCommand around forward/reverse modes 

JohnGilb:
there could be rotational thrashing as the robot was settling on a point and was alternating between tracking a rabbit ahead/behind of it.


JohnGilb:
[https://photos.app.goo.gl/aubiwX5plPv2qLUQ2](url)

JohnGilb:
The solution is to analyze the point the first time you look at it, and "stick" to a forward or reverse rabbit projection.

PR #168, opened 2018-04-04
First thoughts on stall detection module

JohnGilb:


JohnGilb:
Revisit this another time.

PR #187, opened 2019-01-21
On first odometry update, handle non-zero distances

JohnGilb:


JohnGilb:
Tested on robot, looks good.

PR #205, opened 2019-02-10
Pure Pursuit can make 3-point turns

JohnGilb:
The Pure Pursuit command now has another (potentially more efficient) way of navigating the field. Before, due to the motion budget, it had to finish most of its rotation before it began translating.

Now, it translates proportional to the dot product of the robot's vector and the vector to the rabbit. This means that if the rabbit is behind it, it will back up while turning, at 90 degrees it will not translate, and then it will start driving forward as it finishes the turn.

### DIFF
--- a/src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
+++ b/src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
@@ -439,8 +439,10 @@ public abstract class PurePursuitCommand extends BaseCommand {
         // rather than turn in place. And, when we are at 90 degrees to the rabbit, we shouldn't be driving forward
         // or backward at all.
         XYPair robotUnitVector = robot.getHeading().getUnitVector();
+        double dotProduct = robotUnitVector.dotProduct(vectorToRabbit.clone().scale(1/vectorToRabbit.getMagnitude()));
 
         double translationPower = positionalPid.calculate(distanceRemainingToPointAlongPath, 0);
+        translationPower *= dotProduct;
         double constrainedTranslation = constrainToMotionBudget(turnPower, translationPower);
         
         // Log the output. This could be commented out, but for now, it has been very useful for debugging why the robot is driving
--- a/src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
+++ b/src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
@@ -69,6 +69,7 @@ public abstract class PurePursuitCommand extends BaseCommand {
     private int processedPointIndex = -1;
     private boolean stickyPursueForward = true;
     private boolean wasForcingDriveToPoint = false;
+    private boolean useDotProductDriving = false;
 
     /**
      * An implementation of Pure Pursuit for FRC. We got the basic idea from here:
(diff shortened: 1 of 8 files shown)


src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
```diff
@@ -124,6 +125,10 @@ public void setPIDsToDefault() {
         this.headingModule = defaultHeadingModule;
         this.positionalPid = defaultPositionalPid;
     }
+
+    public void setDotProductDrivingEnabled(boolean enabled) {
```
aschokking: I'm wondering if this should be named for the behavior it creates rather than the mechanism by which it achieves that behavior? Like what if we find a better way of doing this than dot-product or maybe some variation thereof?

src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
```diff
@@ -337,10 +343,24 @@ private void advancePointIfNotAtLastPoint() {
      * @param translation The desired translation power, which will be constrained by the budget.
      * @return 
      */
-    private double constrainToMotionBudget(double rotation, double translation) {
-        double remainingBudget = motionBudget.get() - Math.abs(rotation);
-        double constrainedRemainingBudget = MathUtils.constrainDouble(remainingBudget, 0, 1);
-        return translation * constrainedRemainingBudget;
+    private double constrainTranslation(double rotation, double translation, double dotProduct) {
+        if (useDotProductDriving) {
+            return translation *= dotProduct;
```
aschokking: I think it might be worth a comment here about what this does and why it works (range of values of dotProduct and under what circumstances it would be what, 0, 90, 180 etc...)

src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
```diff
@@ -337,10 +343,24 @@ private void advancePointIfNotAtLastPoint() {
...
+        } else {
+            double remainingBudget = motionBudget.get() - Math.abs(rotation);
```
aschokking: In theory now that we're not doing these steps, how is the blending of translation and rotation handled by the drive, is it doing our 'standard' arcade drive math? Eg if it wants to translate at -1.0 for facing the wrong direction but also turn 1.0?

src/main/java/xbot/common/subsystems/drive/PurePursuitCommand.java
```diff
@@ -337,10 +343,24 @@ private void advancePointIfNotAtLastPoint() {
...
+    }
+
+    private double getDotProductWithRabbit(FieldPose robot, XYPair vectorToRabbit, double lookaheadFactor) {
+        // We want to drive towards the rabbit, but what if the rabbit is behind us and we need to rotate?
+        // We should perform a quick 3-point maneuver (back up while rotating, then drive forward while rotating)
+        // rather than turn in place. And, when we are at 90 degrees to the rabbit, we shouldn't be driving forward
+        // or backward at all.
+        XYPair robotUnitVector = robot.getHeading().getUnitVector();
+        return robotUnitVector.dotProduct(
+            vectorToRabbit.clone().scale(1/vectorToRabbit.getMagnitude())) * lookaheadFactor;
```
aschokking: > vectorToRabbit.clone().scale(1/vectorToRabbit.getMagnitude()))

Could this be `vectorToRabbit.getUnitVector()`?

PR #223, opened 2019-09-28
Add settings to use coverage gutters in vscode

aschokking:
VScode settings to display code coverage inline using the Coverage Gutters extension (seems to be the most popular coverage extension at this time).

image


stephenjust:
Maybe add a recommended extensions file?
https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions

PR #226, opened 2019-10-12
Feature/x double solenoid

alyssaki1:
Added a class for double solenoids and tests.


src/main/java/xbot/common/controls/actuators/XDoubleSolenoid.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public void setDoubleSolenoid(DoubleSolenoidMode mode) {
+        if (mode == DoubleSolenoidMode.FORWARD && isInverted)
+        {
+            mode = DoubleSolenoidMode.REVERSE;
+        }
+        else if (mode == DoubleSolenoidMode.REVERSE && isInverted)
+        {
+            mode = DoubleSolenoidMode.FORWARD;
+        }
```
JohnGilb: Here you end up checking for isInverted twice; I'd recommend

```
if (isInverted) {
        if (mode == DoubleSolenoidMode.FORWARD)
        {
            mode = DoubleSolenoidMode.REVERSE;
        }
        else if (mode == DoubleSolenoidMode.REVERSE)
        {
            mode = DoubleSolenoidMode.FORWARD;
        }
}
```

PR #232, opened 2019-10-26
Port RumbleManager code from 2019 competition robot

stephenjust:


stephenjust:
I'm not sure where to actually put the RumbleManager in SeriouslyCommonLib - we don't seem to have any existing place for operator feedback devices.

PR #233, opened 2019-11-02
Prevent solenoids from being created with an invalid channel

stephenjust:
* Added capability to DevicePolice to check channel numbers against a minimum and maximum allowed channel value
* Set WPI solenoid adapter to allow channels from 0-7
* Unit tests for DevicePolice

### DIFF
--- a/src/test/java/xbot/common/injection/wpi_factories/TestDevicePolice.java
+++ b/src/test/java/xbot/common/injection/wpi_factories/TestDevicePolice.java
@@ -1,0 +1,54 @@
+package xbot.common.injection.wpi_factories;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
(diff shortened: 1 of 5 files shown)


src/test/java/xbot/common/injection/wpi_factories/TestDevicePolice.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Test that a device cannot be registered with an id greater than the maximum allowed
+     */
+    @Test(expected = RobotAssertionException.class)
+    public void allocateGreaterThanMax() {
+        RobotAssertionManager ram = injector.getInstance(RobotAssertionManager.class);
+        DevicePolice police = new DevicePolice(ram);
+
+        police.registerDevice(DeviceType.Solenoid, 9000, 0, 7);
```
JohnGilb: :)

PR #272, opened 2020-12-31
DevicePolice: Add device tracking for simulator

aschokking:
- Updates DevicePolice so it can be optionally given the actual device reference (not just the type+id)
- Update XCANTalon to register itself this way
- This will be used by the simulator bridge code to find all the existing devices of a type that have been registered

Longer term I'd rename this class to represent it's broader role but for now this unsticks testing.


src/main/java/xbot/common/controls/actuators/mock_adapters/MockCANTalon.java
```diff
@@ -123,7 +123,11 @@ public void set(ControlMode Mode, double demand0, double demand1) {
     }
 
     public double getThrottlePercent() {
-        return this.throttlePercent;
+        if(master == null){
```
JohnGilb, line 126: Looks like this was fixing a bug - I see that there's some follower logic in the `set()` method, but I guess it wasn't being called properly?

In any case, this seems sensible.

JohnGilb (approved):

PR #274, opened 2021-01-15
Smarter Device Police & SimulationPayloadDistributor

JohnGilb:
Basic idea:

Both sides refer to sensors using their DevicePolice IDs, which are a combination of the kind of port they use (e.g. USB, DigitalIO, CAN, etc..) and the number of the port. The server packages this up into a JSON array of sensors.

The payload is read by a client somewhere, then handed off to the SimulationPayloadDistributor. This reads each sensor entry, gets the full ID, and retrieves the base object.

Each mock sensor will now implement IWebotsSensor (name suggestions welcome), which has one method: `public void  ingestSimulationData(JSONObject payload);` Each actual device will know how to read their specific payload.

Encoders might only need one value - the # of ticks. Our IMU might need a dozen.

### DIFF
--- a/src/main/java/xbot/common/simulation/IWebotsSensor.java
+++ b/src/main/java/xbot/common/simulation/IWebotsSensor.java
@@ -1,0 +1,7 @@
+package xbot.common.simulation;
+
+import org.json.JSONObject;
+
+public interface IWebotsSensor {
+    public void ingestSimulationData(JSONObject payload);
+}
--- a/src/main/java/xbot/common/simulation/SimulationPayloadDistributor.java
+++ b/src/main/java/xbot/common/simulation/SimulationPayloadDistributor.java
@@ -5,31 +5,47 @@ import com.google.inject.Inject;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
+import org.apache.log4j.Logger;
+
 import xbot.common.injection.wpi_factories.DevicePolice;
 
 public class SimulationPayloadDistributor {
 
     private DevicePolice police;
+    private static Logger log = Logger.getLogger(SimulationPayloadDistributor.class);
 
     @Inject
     public SimulationPayloadDistributor(DevicePolice police) {
         this.police = police;
     }
 
+    /**
+     * Fans out data from a simulation (like Webots) to all the simulatable devices registered on the robot.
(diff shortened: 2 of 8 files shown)


src/main/java/xbot/common/simulation/IWebotsSensor.java
```diff
@@ -0,0 +1,7 @@
+package xbot.common.simulation;
+
+import org.json.JSONObject;
+
+public interface IWebotsSensor {
```
aschokking, line 5: Thoughts on this being more generically about simulator than webots? Eg `SimulatableSensor`

src/main/java/xbot/common/simulation/SimulationPayloadDistributor.java
```diff
@@ -0,0 +1,35 @@
...
+    public void distributeSimulationPayload(JSONObject allSensorsPayload) {
+        // For each type underneath
+
+        JSONArray allSensorsArray = (JSONArray)allSensorsPayload.get("Sensors");
+        allSensorsArray.forEach(item -> {
+            JSONObject sensor = (JSONObject)item;
+            String id = (String)sensor.get("ID");
+            JSONObject payload = (JSONObject)sensor.get("Payload");
+
+            System.out.println("ID:"+id+", Value:"+payload);
```
aschokking, line 28: debug logging?

src/main/java/xbot/common/simulation/SimulationPayloadDistributor.java
```diff
@@ -0,0 +1,35 @@
...
+
+            IWebotsSensor device = (IWebotsSensor)police.registeredChannels.get(id);
```
aschokking, line 30: this might be null right?

aschokking (approved):
A few small comments/suggestions

src/main/java/xbot/common/simulation/IWebotsSensor.java
JohnGilb: Good idea - there's nothing in here Webots-specific, so I'll change to a more generic name.

src/main/java/xbot/common/simulation/SimulationPayloadDistributor.java
JohnGilb: Haha, yes, nice catch! I'll clean it up.

PR #282, opened 2021-02-12
Move shared simulator stuff into SCL

aschokking:


src/main/java/xbot/common/command/BaseRobot.java
```diff
@@ -253,4 +267,26 @@ else if(timeSinceLastLog >= frequencyReportInterval.get()) {
         
         outsidePeriodicMonitor.start();
     }
+
+    
+    @Override
```
aschokking, line 272: Azure is complaining about these @Override but they are fine locally

JohnGilb: When I do a clean build of SCL locally, I hit the same error. Something is probably referencing an older version of WPILib.

JohnGilb: Yeah, everything in SCL is pointing to 2020.1.2 versions; those probably need to be incremented.

JohnGilb (approved):

PR #284, opened 2021-02-19
Consume world time from simulator

stephenjust:


src/main/java/xbot/common/controls/sensors/mock_adapters/MockSettableTimer.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.google.inject.Singleton;
+
+import xbot.common.controls.sensors.XSettableTimerImpl;
+
+/**
+ * A settable timer for use in simulations
+ */
+@Singleton
+public class MockSettableTimer implements XSettableTimerImpl {
```
stephenjust, line 11: This is basically exactly the same as MockTimer but with an interface to make the setters accessible. Should I just update MockTimer itself?

JohnGilb: That makes sense - I can't think of a reason we'd want to use the older MockTimer given that the settable one just makes it easier to use when running tests.

JohnGilb (approved):

PR #300, opened 2021-12-04
2022 migration WIP

aschokking:


build.gradle
```diff
@@ -23,61 +28,65 @@ repositories {
...
+    maven { url "https://www.kauailabs.com/maven2" }
+    maven { url "https://www.revrobotics.com/content/sw/max/sdk/maven/"}
     maven { url uri ("../SeriouslyCommonLib/lib/")}
 }
 
 configurations {
     nativeBundle
 }
 
-List<String> wpilibDesktopJni(String platform) {
```
aschokking, line 36: wpi.wpilibVersion doesn't exist anymore, I just pulled all this out and it seemed to be okay but maybe something will burn us later

build.gradle
```diff
@@ -97,27 +106,6 @@ task jarSources(type:Jar){
...
-
-task checkstyle(type: Checkstyle) {
-    source 'src/main/java'
-    configProperties.basedir = file('.')
-    
-    // Empty classpath
-    classpath = files()
-}
-
-checkstyleMain{
```
aschokking, line 113: having these in here caused errors so I removed these 2, there might be a better way to set the checkstyle logging threshold

PR #304, opened 2022-01-15
Move from ContiguousHeading to Rotation2d from WPILib

stephenjust:
Rotation2d is heavily used by WPI's kinematics functions so it's more natural to use it instead of our own implementation.

Add a simple WrappedRotation2d to provide similar value wrapping to old behavior.


src/main/java/xbot/common/math/WrappedRotation2d.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import edu.wpi.first.math.MathUtil;
+import edu.wpi.first.math.geometry.Rotation2d;
+
+/**
+ * A rotation in a 2d coordinate frame, with its rotation wrapped from -pi to pi radians (-180 to 180 degrees).
+ * Prefer this class over Rotation2d when representing robot pose.
```
JohnGilb, line 11: I feel like I must be missing something - what is the purpose of the WrappedRotation2d? All of its constructors seem to have a matching one in [Rotation2d](https://github.com/wpilibsuite/allwpilib/blob/main/wpimath/src/main/java/edu/wpi/first/math/geometry/Rotation2d.java).

stephenjust: Rotation2d does not constrain angles from -180 to 180 degrees, which broke some existing tests that depended on this behavior from ContiguousHeading.

IMO we don't really need the modulus logic except maybe where we report heading to the driver station as a compass direction, but for the initial version I erred on the side of maintaining existing behavior.

If there is agreement that we don't need to normalize these angles, I can remove WrappedRotation2d and update test expectations.

JohnGilb: Ah, I see now - because the constructor uses angleModulus, and the rest of the constructors implicitly call that one. Thanks!

JohnGilb (approved):

PR #313, opened 2022-02-04
Feature/simulate spark max

stephenjust:
* Spark Max simulation
* Maintainer command properties prefixed by subsystem prefix
* Add properties for sim local config

### DIFF
--- a/src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
+++ b/src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
@@ -59,8 +59,6 @@ public abstract class XCANSparkMax {
         prefix = pf.getPrefix();
         policeTicket = police.registerDevice(DeviceType.CAN, deviceId, this);
 
-        setInverted(deviceInfo.inverted);
-
         kPprop = pf.createPersistentProperty("kP", defaultPIDProperties.p);
         kIprop = pf.createPersistentProperty("kI", defaultPIDProperties.i);
         kDprop = pf.createPersistentProperty("kD", defaultPIDProperties.d);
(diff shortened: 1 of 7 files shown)


src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
```diff
@@ -38,23 +39,27 @@
...
-        this.deviceId = deviceId;
+        this.deviceId = deviceInfo.channel;
         this.propertyFactory = pf;
         this.propertyFactory.setPrefix(owningSystemPrefix);
         this.propertyFactory.appendPrefix(name);
         prefix = pf.getPrefix();
-        police.registerDevice(DeviceType.CAN, deviceId, this);
+        policeTicket = police.registerDevice(DeviceType.CAN, deviceId, this);
+
+        setInverted(deviceInfo.inverted);
```
JohnGilb, line 62: This has to be done in some kind of delayed fashion - if we do this in the constructor, it will attempt to call `setInverted` on CANSparkMaxWpiAdapter, which will attempt to call `internalSpark.setInverted(isInverted)`... but the internalSpark won't have been created yet!

We didn't have a great solution for this before - you might notice that XCANTalon has to deal with the same issue, and it "solves" it by adding some extra configuration methods like `configureMotorTeam` and `configureAsMasterMotor` that are meant to be called after the object is created. There may be a better way to do this, though. Maybe the abstract class sets some flag that the subclass reads during its own constructor?

JohnGilb (changes requested):

JohnGilb (approved):

src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
stephenjust: CANTalonWpiAdapter just sets inverted after assigning the internal object. I updated this to follow the same pattern.

PR #353, opened 2022-10-16
Fix generation of Dagger outputs for non-gradle builds

stephenjust:
When VS Code builds java code for intellisense / running tests / debugging, it doesn't use Gradle directly. The VS Code java language plugin internally uses Eclipse's Java build tools to try to run a gradle project.

Essentially this change tells the Gradle build to generate eclipse files (.classpath, .factorypath, etc) when run directly, and the VS Code-triggered build (for tests or debugging) can also consume the plugin outputs. The VS Code Java language server also uses this classpath to search for source files (for intellisense). The additional plugins let the Eclipse-driven build also run dagger code generation, which means that intellisense should just work.

Classpath and factorypath are generated at build and not checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ WPI_Native_Libraries.properties
 #Gradle
 .gradle/
 
+#Eclipse (build tools used by VS Code)
 .settings/
 .classpath
 .factorypath
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,6 @@ WPI_Native_Libraries.properties
 
 .settings/
 .classpath
+.factorypath
 .project
 *.dll
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ import edu.wpi.first.gradlerio.deploy.roborio.RoboRIO
  */
 
 plugins {
-    id "edu.wpi.first.GradleRIO" version "2022.3.1"
+    id "edu.wpi.first.GradleRIO" version "2022.4.1"
     id "eclipse"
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -159,6 +161,26 @@ tasks.withType(Test) {
     }
 }
 
+eclipse {
+	// re-generate APT configuration on Gradle project synchronisation
+	synchronizationTasks eclipseJdtApt, eclipseJdt, eclipseFactorypath
+
+	jdt.apt {
+        genSrcDir = file('bin/generated/sources/annotationProcessor/java/main')
```
aschokking, line 169: ❔  just curious, why switch from build to bin?

.gitignore
```diff
@@ -37,5 +37,6 @@ WPI_Native_Libraries.properties
 
 .settings/
 .classpath
+.factorypath
```
aschokking, line 40: is this from the eclipse plugin? consider adding a comment

aschokking (approved):
Thanks for digging into this hairy area! I'm excited to see if this fixes the edu vs code issues

build.gradle
stephenjust: There's a convention of keeping eclipse and gradle build outputs separate. Bin is the eclipse path, we don't need to add the gradle path (in build), since classpath is an eclipse toolchain thing.

.gitignore
stephenjust: Yes all of these three are from the eclipse build tool

PR #367, opened 2023-02-07
Integrate AdvantageKit and Log4j2

JohnGilb:
- Integrates AdvantageKit
  - Robot now explicitly goes through three phases - Refresh Data, Process Data, and Execute Commands
  - Ephemeral properties are functionally deprecated, being replaced by calls to log data to disk. (Anything recorded in AdvantageKit is also replicated as a NetworkTable value, so it's still discoverable at runtime with Shuffleboard or AdvantageScope.)
  - Any system that intermittently logged data to the console instead just writes directly to the log
  - Concepts like a logging session are now managed by AdvantageKit
- Removes Log4j, Adds Log4j2.
  - Very little changes here, except that calls to get the logging object are of the form `Logger log = LogManager.getLogger(name)`
  - Sadly, this does change a billion files.


stephenjust (approved):
😲 That's a big diff

build.gradle
```diff
@@ -29,6 +29,20 @@ repositories {
     maven { url "https://mvnrepository.com/artifact/org.zeromq/jeromq" }
 
     maven { url uri ("../SeriouslyCommonLib/lib/")}
+
+    maven {
+        url = uri("https://maven.pkg.github.com/Mechanical-Advantage/AdvantageKit")
+        credentials {
```
aschokking, line 35: ⭐  can you add a comment about where these came from in case we ever have to update them?

src/main/java/xbot/common/command/BaseMaintainerCommand.java
```diff
@@ -11,7 +11,7 @@
 
 public abstract class BaseMaintainerCommand<T> extends BaseCommand {
 
-    BaseSetpointSubsystem subsystemToMaintan;
+    BaseSetpointSubsystem subsystemToMaintain;
```
aschokking, line 14: where's the fun in that 🤣

src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
```diff
@@ -19,24 +24,31 @@ public abstract class XCANSparkMax {
...
     PropertyFactory propertyFactory;
 
-    final DoubleProperty kPprop;
-    final DoubleProperty kIprop;
-    final DoubleProperty kDprop;
-    final DoubleProperty kIzProp;
-    final DoubleProperty kFFprop;
-    final DoubleProperty kMaxOutputProp;
-    final DoubleProperty kMinOutoutProp;
+    private boolean usesPropertySystem = true;
```
aschokking, line 30: I've been thinking in general we probably should keep things protected rather than private so that extension is more possible?

aschokking (approved):

PR #390, opened 2023-10-21
Integrate AdvantageKit into SCL

JohnGilb:
Brings in core AdvantageKit functionality to the SCL.


src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,13 @@
...
+         <PatternLayout pattern="%5p | %d | %L | %-40c{1} | %m%n"/>
+      </Console>
+   </Appenders>
+   <Loggers>
+      <Logger name="org.apache.log4j.xml" level="info"/>
+      <Root level="info">
+         <AppenderRef ref="STDOUT"/>
+      </Root>
+   </Loggers>
+</Configuration>
```
stephenjust, line 13: Try this to restore the file appender as well:
```xml
<Configuration>
   <Properties>
      <Property name="filePattern">488Logs/default</Property>
   </Properties>
   <Appenders>
      <Console name="STDOUT" target="SYSTEM_OUT">
         <PatternLayout pattern="%5p | %d | %L | %-40c{1} | %m%n"/>
      </Console>
      <RollingFile name="File" fileName="${filePattern}.log" filePattern="${filePattern}-%i.log">
         <PatternLayout pattern="%5p | %d | %L | %-40c{1} | %m%n"/>
         <Policies>
            <OnStartupTriggeringPolicy/>
         </Policies>
         <DefaultRolloverStrategy max="10"/>
      </RollingFile>
   </Appenders>
   <Loggers>
      <Logger name="org.apache.log4j.xml" level="info"/>
      <Root level="debug">
         <AppenderRef ref="STDOUT"/>
         <AppenderRef ref="File"/>
      </Root>
   </Loggers>
</Configuration>
```

JohnGilb: This is me remembering from early 2023, but I think this was intentional since AdvantageKit is already handling all the "writing to disk" that the FileAppender would have done.

src/main/java/xbot/common/command/BaseMaintainerCommand.java
```diff
@@ -126,13 +120,14 @@ protected void initializeMachineControlAction() {
...
-        // is trying to manipulate the setpoint.
+        // is trying to maniuplate the setpoint.
+
         if (subsystemToMaintain.getSetpointLock().getCurrentCommand() != null && !DriverStation.isAutonomous()) {
             subsystemToMaintain.getSetpointLock().getCurrentCommand().cancel();
         }
 
         // Typically set the goal to the current position, to avoid sudden extreme
-        // changes
+        // changes`
```
aschokking, line 130: typo?

src/main/java/xbot/common/command/BaseRobot.java
```diff
@@ -117,6 +91,22 @@ protected BaseComponent getInjectorComponent() {
      * This function is run when the robot is first started up and should be used for any initialization code.
      */
     public void robotInit() {
+
+        Logger.getInstance().recordMetadata("ProjectName", "XbotProject"); // Set a metadata value
+        if (isReal() || forceWebots) {
+            Logger.getInstance().addDataReceiver(new WPILOGWriter("/media/sda1/")); // Log to a USB stick
```
aschokking, line 97: do we need to guard against the stick being missing or is this pretty tolerant?

src/main/java/xbot/common/command/BaseRobot.java
```diff
@@ -252,62 +237,63 @@ public void teleopPeriodic() {
...
 
     
     @Override
     public void simulationInit() {
-        /*
-        webots = injectorComponent.webotsClient();
-        webots.initialize();
-
-         */
+        // TODO: Add something to detect replay vs Webots, and skip all of this if we're in replay mode.
```
aschokking, line 265: not for now, but instead of `if webots` maybe should have an enum of simutation modes?

src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
```diff
@@ -739,4 +774,35 @@ public abstract REVLibError setReference(double value, ControlType ctrl, int pid
     public abstract boolean getForwardLimitSwitchPressed(com.revrobotics.SparkLimitSwitch.Type switchType);
 
     public abstract boolean getReverseLimitSwitchPressed(com.revrobotics.SparkLimitSwitch.Type switchType);
+
+    public abstract REVLibError setPeriodicFramePeriod(CANSparkLowLevel.PeriodicFrame frame, int periodMs);
+
+    /**
+     * // See https://docs.revrobotics.com/sparkmax/operating-modes/control-interfaces#periodic-status-frames
```
aschokking, line 781: It might be worth adding more 1 tldr here about what status frames are, something along the lines of the tail end of this comment about pid?

src/main/java/xbot/common/controls/actuators/wpi_adapters/CANSparkMaxWpiAdapter.java
```diff
@@ -258,19 +259,21 @@ public boolean getFault(FaultID faultID) {
...
 
-    public boolean getStickyFault(FaultID faultID) {
+    public boolean getStickyFault(FaultID faultID) { return internalSpark.getStickyFault(faultID); }
+
+    public boolean getStickyFault_internal(FaultID faultID) {
         return internalSpark.getStickyFault(faultID);
     }
 
-    public double getBusVoltage() {
+    public double getBusVoltage_internal() {
```
aschokking, line 268: is this underscore naming coming from the can spark max interface?

src/main/java/xbot/common/logging/WordGenerator.java
aschokking, line 1: thank you for your service

aschokking (approved):
Wooo there's a lot here, excited to try it out on the robo

src/main/java/xbot/common/command/BaseRobot.java
JohnGilb: From my memory last year the robot wouldn't crash; it would just fail to log. But should be revisited.

src/main/java/xbot/common/controls/actuators/wpi_adapters/CANSparkMaxWpiAdapter.java
JohnGilb: No - this is me trying to keep most things backwards compatible, but a better name would be welcome. The basic idea is as such:

* The `XCANSparkMax` abstract class exposes the "simple" name like `getPosition()`, since that's what subsystems/etc. will be calling.
* The implementations like `CANSparkMaxWpiAdapter` need to immediately return the info, but per AdvantageKit philosophy, they can't contact the underlying object; they instead need to refer to a cached value. So `getPosition()` just calls `return inputs.position`.
* At some point we do actually need to call the underlying object to fill that cache, and that method is `getPosition_internal()`, called in `updateInputs()`

PR #391, opened 2023-10-21
Update to WPILib 2024

stephenjust:


src/main/java/xbot/common/controls/actuators/wpi_adapters/PWMWPIAdapter.java
```diff
@@ -10,7 +10,7 @@
 
 public class PWMWPIAdapter extends XPWM
 {
-    private PWM pwm;
+    private final PWM pwm;
```
aschokking, line 13: it's a good idea, but just curious why it was needed?

aschokking (approved):
Looks good, thanks for working through this so fast!

src/main/java/xbot/common/controls/actuators/wpi_adapters/PWMWPIAdapter.java
stephenjust: Since I was changing the file I just fixed the warnings while I was at it.

PR #406, opened 2024-01-17
Integrates PhotonVision objects with AdvantageKit

JohnGilb:
* Creates PhotonCameraExtended, which logs information to Akit.
* Delete other ones


stephenjust:
Do we really need the forked components anymore? It looks like you should be ablel to access what you need with inheritance now?

stephenjust (approved):

PR #414, opened 2024-01-26
Silence warnings about deprecated CANCoder class

stephenjust:
# Why are we doing this?
Warnings during build show up as red text and are confusing to students.

# Whats changing?
Deprecation warnings are silenced.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking (approved):
I'm all for silencing warnings we won't do anything about.

Was just curious if there was a way to share this logic so we didn't need to repeat it in as many places?

stephenjust:
> I'm all for silencing warnings we won't do anything about.
> 
> Was just curious if there was a way to share this logic so we didn't need to repeat it in as many places?

I could not find a way to share, no.

PR #417, opened 2024-01-26
Support absolute encoders attached to SparkMax objects

JohnGilb:
# Why are we doing this?
Current plans for the robot involve attaching a through-bore encoder to the Arm assembly, and plugging that directly into a nearby SparkMax rather than the RIO.

# What's changing?
XCANSparkMax objects can be asked for a XSparkAbsoluteEncoder object, which in turn must be refreshed like any other base sensor.

I put this together to unblock a student working in the area, but there's probably some way to either join this closer to the existing XAbsoluteEncoder or more strongly separate them.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
+++ b/src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
@@ -8,12 +8,14 @@ import com.revrobotics.CANSparkBase.SoftLimitDirection;
 import com.revrobotics.CANSparkLowLevel;
 import com.revrobotics.CANSparkMax;
 import com.revrobotics.REVLibError;
+import com.revrobotics.SparkAbsoluteEncoder;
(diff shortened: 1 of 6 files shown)


src/main/java/xbot/common/controls/actuators/XCANSparkMax.java
```diff
@@ -790,6 +792,8 @@ public void setupStatusFramesIfReset(int status0PeriodMs, int status1PeriodMs, i
             }
     }
 
+    public abstract XSparkAbsoluteEncoder getAbsoluteEncoder(String nameWithPrefix, boolean inverted);
```
aschokking, line 795: ⭐ consider a little comment here about what situation this is for

aschokking (approved):
sounds reasonable to me

PR #418, opened 2024-01-27
Properties now logged by AKIT

JohnGilb:
# Why are we doing this?
* We should add configuration properties to AKIT, so it can 100% replay the match from the logged data. However, this would cause them to show up in 3 places - Preferences, SmartDashboard, and AKIT. That's a little excessive
* The Preferences system already handles using a network table for random access, so our save/load mechanism was no longer relevant - it can be retired.
# Whats changing?
* Ephemeral properties are gone - replaced by calls to the AKIT logger.
* Double/String/BooleanProperties now log and hydrate from AKIT like other systems
* The PropertyManager refreshes all properties at the top of each cycle.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


stephenjust (changes requested):
Can you create some documentation for the new expected usage pattern?

There should be a package-info.java in the Properties package where you can put sample code in the javadoc comment.

(Use \<pre\>\</pre\> to preserve code formatting, preview with `./gradlew.bat javadoc` and review the output in build/doc/javadoc/index.html.)

stephenjust (approved):

PR #420, opened 2024-01-27
Fix Digital Input bug

JohnGilb:
# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/controls/sensors/XDigitalInput.java
```diff
@@ -21,6 +21,7 @@ public interface XDigitalInputFactory {
     public XDigitalInput(DevicePolice police, DeviceInfo info) {
         police.registerDevice(DeviceType.DigitalIO, info.channel, this);
         inputs = new XDigitalInputsAutoLogged();
+        this.info = info;
```
aschokking, line 24: Should we mark these final so the bug would be impossible?

aschokking (approved):

PR #423, opened 2024-01-27
AKitLogger

aschokking:
# Why are we doing this?
We're using advantage kit more and more to do what we used to use ephemeral property logging. The new syntax is a little verbose though with us having to do prefix setting ourselves.

# Whats changing?

Adds a new logger class that wraps akit and does our prefixing.

```diff
-        Logger.getInstance().recordOutput(this.getPrefix() + "ErrorWithinTotalTolerance", withinErrorTolerance);
-        Logger.getInstance().recordOutput(this.getPrefix() + "ErrorIsTimeStable", isStable);
+        aKitLog.record("ErrorWithinTotalTolerance", withinErrorTolerance);
+        aKitLog.record("ErrorIsTimeStable", isStable);
```

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206445513269033

### DIFF
--- a/src/main/java/xbot/common/advantage/AKitLogger.java
+++ b/src/main/java/xbot/common/advantage/AKitLogger.java
@@ -1,0 +1,113 @@
+package xbot.common.advantage;
+
+import org.littletonrobotics.junction.Logger;
+
+import edu.wpi.first.units.Measure;
+import edu.wpi.first.units.Unit;
+import edu.wpi.first.util.WPISerializable;
+import edu.wpi.first.util.protobuf.Protobuf;
+import edu.wpi.first.util.struct.Struct;
+import edu.wpi.first.util.struct.StructSerializable;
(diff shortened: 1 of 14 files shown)


JohnGilb (approved):
This is really nice! For the AKitLogger itself, did you write that or find some utility to setup most of the delegate methods? (Thinking about if we get an AdvantageKit update that adds/removes a dozen methods, is there an approach other than manual writing)

aschokking:
> This is really nice! For the AKitLogger itself, did you write that or find some utility to setup most of the delegate methods? (Thinking about if we get an AdvantageKit update that adds/removes a dozen methods, is there an approach other than manual writing)

I used chat gpt, I gave it one example of the transform I wanted and it did the rest

PR #425, opened 2024-01-28
Pipe in basic velocity control of drive wheels

JohnGilb:
# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/command/BaseMaintainerCommand.java
+++ b/src/main/java/xbot/common/command/BaseMaintainerCommand.java
@@ -207,11 +207,6 @@ public abstract class BaseMaintainerCommand<T> extends BaseCommand {
      */
     protected abstract double getHumanInputMagnitude();
 
-    @Override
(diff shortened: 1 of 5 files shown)


src/main/java/xbot/common/command/BaseMaintainerCommand.java
```diff
@@ -207,11 +207,6 @@ protected boolean getErrorWithinTolerance() {
      */
     protected abstract double getHumanInputMagnitude();
 
-    @Override
-    public String getPrefix() {
-        return subsystemToMaintain.getPrefix() + getName() + "/";
```
JohnGilb, line 212: This just fixes a nullref, since the maintainer command will be created before the subsystem to maintain is set. On the upside, at least the records for this class will be just like other commands.

aschokking (approved):

PR #426, opened 2024-01-28
Feature/revisit dynamic pathing

JohnGilb:
# Why are we doing this?
The existing pathing logic was a bit old and not well documented; updated and fixed several bugs.
# Whats changing?
* Lots of bugfixes related to interesting collision types, particularly when the ray would cross opposite sides of the obstacle.
* If the robot gets really disconnected from the ghost (collision, possibly human intervention), it can recover and get back on the line of travel.

# How this was tested
- [x] unit tests added
- [x] tested on robot
- [x] tested in local simulator

### DIFF
--- a/src/main/java/xbot/common/command/BaseMaintainerCommand.java
+++ b/src/main/java/xbot/common/command/BaseMaintainerCommand.java
@@ -207,11 +207,6 @@ public abstract class BaseMaintainerCommand<T> extends BaseCommand {
      */
     protected abstract double getHumanInputMagnitude();
 
-    @Override
-    public String getPrefix() {
-        return subsystemToMaintain.getPrefix() + getName() + "/";
-    }
-
     /**
      * Sets the error tolerance for the maintain command.
      * @param tolerance The error tolerance.
--- a/src/main/java/xbot/common/trajectory/LowResField.java
+++ b/src/main/java/xbot/common/trajectory/LowResField.java
@@ -23,6 +23,12 @@ public class LowResField {
         obstacles = new ArrayList<Obstacle>();
     }
 
+    /**
+     * Adds an obstacle to the field. If the obstacle intersects with any existing obstacles, it will be combined with
+     * the existing obstacle into a single larger obstacle with a bounding box that fully contains both obstacles.
+     * This should be the only method used to add new obstacles; otherwise obstacles could intersect.
+     * @param o The obstacle to add.
+     */
     public void addObstacle(Obstacle o) {
         Obstacle potentialObstacleToRemove = null;
         for (Obstacle existingObstacle : obstacles) {
@@ -47,13 +53,6 @@ public class LowResField {
         return new ArrayList<Obstacle>(obstacles);
     }
 
-    public void squashObstacles() {
-        // Check if any obstacles are inside other obstacles.
-        // If so, combine them into a new obstacle that fully contains the previous two.
-
-
-    }
-
     /**
      * Generates a path from the current position to the target point, avoiding known obstacles.
      * Very generally, it will recursively solve paths nearby the target, and then step foward to approach the target.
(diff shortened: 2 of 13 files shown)


src/main/java/xbot/common/command/BaseMaintainerCommand.java
```diff
@@ -207,11 +207,6 @@ protected boolean getErrorWithinTolerance() {
      */
     protected abstract double getHumanInputMagnitude();
 
-    @Override
-    public String getPrefix() {
-        return subsystemToMaintain.getPrefix() + getName() + "/";
```
JohnGilb, line 212: This causes a nullref as the subsystem can't easily be created. Maintainer records will need to live under their own command, which at least matches up with other commands.

src/main/java/xbot/common/trajectory/LowResField.java
```diff
@@ -24,13 +24,306 @@ public LowResField() {
...
+        Translation2d currentSource = currentPose.getTranslation();
+        //      b) initialize an empty list of swerve points.
+        List<XbotSwervePoint> path = new ArrayList<XbotSwervePoint>();
+        // 2) Create a Stack of XbotSwervePoints, which will be used to help us deal with multiple intermediate points.
+        Stack<XbotSwervePoint> swervePointStack = new Stack<XbotSwervePoint>();
+
+        XbotSwervePoint specialFinalPoint = null;
+
+        for (Obstacle o : obstacles) {
+            o.resetCorners();
```
aschokking, line 107: it feels a little icky to me for this function to have side effects on obstacles, maybe this should be working on a copy of them?

src/main/java/xbot/common/trajectory/LowResField.java
```diff
@@ -24,13 +24,306 @@ public LowResField() {
...
+                    if (specialFinalPoint != null) {
+                        path.add(specialFinalPoint);
+                    }
+
+                    completed = true;
+                }
+            }
+
+        }
+        return path;
```
aschokking, line 324: this is a pretty enormous function (not just from the comments), what do you think about breaking it down into a few smaller pieces so the outer function reads a little easier?

I also wonder if those inner parts wouldn't be easier to unit test as well?

src/test/java/xbot/common/trajectory/LowResFieldTest.java
```diff
@@ -57,6 +59,7 @@ private FieldPose flipFieldPose(FieldPose current) {
                 new WrappedRotation2d(180 - current.getHeading().getDegrees()));
     }
 
+    @Ignore
```
aschokking, line 62: ⭐ could you add a comment about why these are ignored? Is it a TODO to get them passing again?
eg:
`@Ignore("Broken from change X, TODO to get passing again.")`

aschokking (approved):
Wooh there's a lot here! I don't 100% follow all of the logic but have some readability suggestions

src/main/java/xbot/common/trajectory/LowResField.java
```diff
@@ -24,13 +24,306 @@ public LowResField() {
...
 
     public void addObstacle(Obstacle o) {
-        obstacles.add(o);
+        Obstacle potentialObstacleToRemove = null;
+        for (Obstacle existingObstacle : obstacles) {
+            if (existingObstacle.intersects(o) || existingObstacle.contains(o) || o.contains(existingObstacle)) {
+                log.info("Collides with existing obstacle. Combining.");
+                existingObstacle.absorbObstacle(o);
+                potentialObstacleToRemove = existingObstacle;
+                break;
```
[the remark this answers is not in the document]
JohnGilb: Yes; let me add that as documentation.

src/main/java/xbot/common/trajectory/LowResField.java
```diff
@@ -24,13 +24,306 @@ public LowResField() {
...
+    }
+
+    /**
+     * Generates a path from the current position to the target point, avoiding known obstacles.
+     * Very generally, it will recursively solve paths nearby the target, and then step foward to approach the target.
+     * @param currentPose Starting point
+     * @param targetPoint Terminating Point
+     * @return A List of XbotSwervePoints that can be followed to reach the destination without collision.
+     */
+    public List<XbotSwervePoint> generatePathFrontToBack(Pose2d currentPose, XbotSwervePoint targetPoint) {
```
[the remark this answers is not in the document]
JohnGilb: Good point - after some testing, I think the new method can take the mantle of `generatePath()` and we can retire the old one.

src/main/java/xbot/common/trajectory/LowResField.java
```diff
@@ -24,13 +24,306 @@ public LowResField() {
...
+        // New attempt at a pathfinding algorithm that maybe a little more resilient. The original algorithm had
```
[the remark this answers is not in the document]
JohnGilb: original obliterated

PR #428, opened 2024-02-02
BaseMaintainerCommand doc

aschokking:
# Why are we doing this?
It can be hard to understand this maintainer / decision system.

# Whats changing?
Adding a _little_ bit more documentation and a state flow diagram

### DIFF
--- a/src/main/java/xbot/common/command/BaseMaintainerCommand.java
+++ b/src/main/java/xbot/common/command/BaseMaintainerCommand.java
@@ -1,20 +1,32 @@
 package xbot.common.command;
 
-import org.littletonrobotics.junction.Logger;
 import edu.wpi.first.wpilibj.DriverStation;
 import xbot.common.logic.HumanVsMachineDecider;
 import xbot.common.logic.HumanVsMachineDecider.HumanVsMachineDeciderFactory;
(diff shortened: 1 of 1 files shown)


JohnGilb:
I'm still learning about Javadoc; should this be in some kind of package-info rather than on one class? Or is there an easy way for somebody to view this directly in their IDE?

JohnGilb (approved):

aschokking:
> I'm still learning about Javadoc; should this be in some kind of package-info rather than on one class? Or is there an easy way for somebody to view this directly in their IDE?

I only know about how to see it in the browser, eg for this class: https://team488.github.io/SeriouslyCommonLib/xbot/common/command/BaseMaintainerCommand.html

stephenjust:
> > I'm still learning about Javadoc; should this be in some kind of package-info rather than on one class? Or is there an easy way for somebody to view this directly in their IDE?
> 
> I only know about how to see it in the browser, eg for this class: https://team488.github.io/SeriouslyCommonLib/xbot/common/command/BaseMaintainerCommand.html

There's nothing in the IDE that will render mermaid, but the javadoc will show up if you hover over the class in the IDE. For package-info, it'll show up when you hover over the package name in your imports.

PR #434, opened 2024-02-09
CanCoder devices should always be in -180 to 180 degrees mode since t…

stephenjust:
…his is the WPI standard way of handling angles

# Why are we doing this?
This avoids needing to remember to configure this setting when preparing new CANCoders

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking (approved):
❤️

PR #446, opened 2024-02-23
Add log level concept to AKitLogger

aschokking:
# Why are we doing this?
We're starting to have too many values being sent to akit. This should allow us to mark some as debug and not send them during competitions at least.

https://app.asana.com/0/38541457243752/1206617786624119

# Whats changing?
- new static log level on AKitLogger that would be set to INFO during competitions (statically, maybe in Robot.java?) and otherwise DEBUG
- individual loggers can be set into debug or info mode before calling to log. If the global level is info but the local level is debug, the call to log will no-op

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/advantage/AKitLogger.java
+++ b/src/main/java/xbot/common/advantage/AKitLogger.java
@@ -13,8 +13,23 @@ import us.hebi.quickbuf.ProtoMessage;
 import xbot.common.properties.IPropertySupport;
 
 public class AKitLogger {
+    public enum LogLevel {
+        DEBUG, INFO
+    }
+    private static LogLevel globalLogLevel = LogLevel.DEBUG;
 
     private String prefix = "";
+    private LogLevel logLevel = globalLogLevel;
+
+    /**
+     * This controls the log level for all AKitLoggers. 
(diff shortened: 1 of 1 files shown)


src/main/java/xbot/common/advantage/AKitLogger.java
```diff
@@ -24,91 +39,164 @@ public AKitLogger(IPropertySupport parent) {
...
+     * Log calls made after this will have that level when checking 
+     * if they should record or not.
+     * @param level new level to set
+     */
+    public void setLogLevel(LogLevel level) {
+        this.logLevel = level;
+    }
+
     public void record(String key, byte[] value) {
+        if(this.logLevel == LogLevel.DEBUG && globalLogLevel == LogLevel.INFO) {
```
JohnGilb, line 53: What about a function, `private boolean shouldLog()` so each of these lines can just be `if(shouldLog())` ?

JohnGilb (approved):

src/main/java/xbot/common/advantage/AKitLogger.java
aschokking: Yup good point, I was stuck on the return not being something that could be refactored away

PR #473, opened 2024-03-21
Gamepad vector updates

aschokking:
# Why are we doing this?

Our gamepad logic was still using XYPair which we're moving away from in favor of the wpilib math stuff. While here I also added the ability to get a field oriented vector from the gamepad.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/controls/sensors/XFTCGamepad.java
```diff
@@ -44,18 +46,38 @@ protected int getRightTriggerAxis() {
...
     }
 
-    public XYPair getRightVector() {
+    public Translation2d getRightVector() {
         return this.getVectorForAxisPair(
                 getRightJoystickXAxis(),
                 getRightJoystickYAxis());
     }
 
+    public Translation2d getLeftFieldOrientedVector() {
```
JohnGilb, line 61: :star: Maybe a function used by both of these? Lots of duplicated code.

JohnGilb (approved):

stephenjust (approved):

JohnGilb:
Converting to draft to further avoid merge chances

PR #478, opened 2024-03-30
remove the dirty flag and use programname instead

rzhai2:
# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/subsystems/autonomous/AutonomousCommandSelector.java
+++ b/src/main/java/xbot/common/subsystems/autonomous/AutonomousCommandSelector.java
@@ -22,13 +22,11 @@ public class AutonomousCommandSelector extends BaseSubsystem {
     Supplier<Command> commandSupplier;
 
     Command currentAutonomousCommand;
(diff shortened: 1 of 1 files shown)


src/main/java/xbot/common/subsystems/autonomous/AutonomousCommandSelector.java
```diff
@@ -63,11 +60,7 @@ public Command createAutonomousStateMessageCommand(String message) {
         });
     }
 
-    public void setIsDefault(boolean bol) {
-        this.isDefault = bol;
-    }
-
-    public boolean getIsDefault() {
-        return this.isDefault;
+    public String getProgramName() {
```
aschokking, line 63: Is there a reason not to leave both so we could debug the default stuff?

One of the big reasons to go with the default approach was that we could signal if something like the Oracle being default had had any settings changed on it (which this getName() approach couldn't handle).

aschokking (approved):

PR #487, opened 2025-01-11
[Feature] Write a universal motor controller interface

stephenjust:
In 2025, we need to be able to control both a Spark and TalonFX powered drive-train. Since we had to re-write the Spark MAX driver anyway, implement both Spark MAX and Talon FX using the same interface.

There are plenty of missing features compared to the previous abstraction, but this should be enough for a swerve chassis.

# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/controls/actuators/mock_adapters/MockCANMotorController.java
```diff
@@ -0,0 +1,95 @@
...
+
+    }
+
+    @Override
+    public void setPidProperties(double p, double i, double d, double velocityFF, int slot) {
+
+    }
+
+    @Override
+    public void setPower(double power) {
```
aschokking, line 59: ⭐ something we can do later, but it would be nice if these stored the last things set so it could be read back out by unit tests/physics sims

src/main/java/xbot/common/injection/DevicePolice.java
```diff
@@ -91,6 +91,37 @@ public String registerDevice(DeviceType type, int id, Object device) {
...
+     * @param busId Device CAN bus id
+     * @param id Device id
+     * @param device Device object
+     */
+    public String registerDevice(DeviceType type, CANBusId busId, int id, Object device) {
+        // First, check to see if the overall device has already been registered once. We only
+        // want there to be one "main" entry for a given device, like an Encoder, which may use two channels.
+        // That way, when we ask the DevicePolice how many devices have been registered, it wouldn't return the
+        // same Encoder twice.
+        if (!registeredDevices.contains(device)) {
```
aschokking, line 106: ⭐ since this check happens in the parent registerDevice call already, why not remove this from here and move the call to this method down in the other call after the check?

src/main/java/xbot/common/injection/DevicePolice.java
```diff
@@ -91,6 +91,37 @@ public String registerDevice(DeviceType type, int id, Object device) {
...
+            registeredDevices.add(device);
+        }
+
+        if (type != DeviceType.CAN) {
```
aschokking, line 110: ⭐ should this be at the top of the method instead of after the other stuff happens?

src/main/java/xbot/common/injection/components/BaseComponent.java
```diff
@@ -166,14 +166,14 @@ public abstract class BaseComponent {
...
 
-    public abstract XCANSparkMaxFactory canSparkMaxFactory();
-
     public abstract XLidarLiteFactory lidarLiteFactory();
 
     public abstract XSpeedControllerFactory speedControllerFactory();
 
     public abstract XZeromqListenerFactory zeromqListenerFactory();
 
+    public abstract XCANMotorController.XCANMotorControllerFactory motorControllerFactory();
```
aschokking, line 175: ⭐ to align with the type, maybe this should be called `canMotorControllerFactory();`?

src/main/java/xbot/common/injection/electrical_contract/CANMotorControllerOutputConfig.java
```diff
@@ -0,0 +1,38 @@
+package xbot.common.injection.electrical_contract;
+
+import edu.wpi.first.units.measure.Current;
+
+import static edu.wpi.first.units.Units.Amps;
+
+public class CANMotorControllerOutputConfig {
+    public enum InversionType {
```
aschokking, line 8: ❔ out of curiosity, why an enum vs a bool?

src/main/java/xbot/common/injection/electrical_contract/CANMotorControllerOutputConfig.java
```diff
@@ -0,0 +1,38 @@
...
+    public enum NeutralMode {
+        Brake,
+        Coast
+    }
+
+    public InversionType inversionType = InversionType.Normal;
+
+    public NeutralMode neutralMode = NeutralMode.Coast;
+
+    public Current statorCurrentLimit = Amps.of(80);
```
aschokking, line 22: ⭐ suggest pulling out this `80` magic number into a named const `DEFAULT_CURRENT_LIMIT`

src/test/java/xbot/common/injection/factories/TestAllFactoryClasses.java
```diff
@@ -50,10 +63,10 @@ public void makeOneOfEverything() {
         getInjectorComponent().doubleSolenoidFactory().create(
             getInjectorComponent().solenoidFactory().create(2),
             getInjectorComponent().solenoidFactory().create(3));
-        getInjectorComponent().canSparkMaxFactory().create(new DeviceInfo("left", 10), "drive", "left", "motorGroup");
-        getInjectorComponent().canSparkMaxFactory().create(
-                new DeviceInfo("left", 11), "drive", "left", "motorGroup",
-                new XCANSparkMaxPIDProperties(1, 0, 0, 0, 0, 0.5, -0.5));
+//        getInjectorComponent().canSparkMaxFactory().create(new DeviceInfo("left", 10), "drive", "left", "motorGroup");
```
aschokking, line 66: ⭐ just delete these? or this is a TODO?

aschokking (commented):
Great job, so much going on here!

PR #489, opened 2025-01-15
"AKIT-ify" XCANMotorController

JohnGilb:
# Why are we doing this?
The XCANMotorController skipped the input processing & `refreshDataFrame()` aspects that get its data recorded by AKIT.
# Whats changing?
- Adds basic XCANMotorControllerInputs, just handling positions and velocity
- Fixes a bug in the naming scheme for motors - steering and drive were using the same functions

# How this was tested
- [x] tested in simulation

### DIFF
--- a/src/main/java/xbot/common/controls/io_inputs/XCANMotorControllerInputs.java
+++ b/src/main/java/xbot/common/controls/io_inputs/XCANMotorControllerInputs.java
@@ -1,0 +1,11 @@
+package xbot.common.controls.io_inputs;
+
+import edu.wpi.first.units.measure.Angle;
+import edu.wpi.first.units.measure.AngularVelocity;
+import org.littletonrobotics.junction.AutoLog;
+
+@AutoLog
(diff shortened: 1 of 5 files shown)


src/main/java/xbot/common/controls/io_inputs/XCANMotorControllerInputs.java
```diff
@@ -0,0 +1,11 @@
+package xbot.common.controls.io_inputs;
+
+import edu.wpi.first.units.measure.Angle;
+import edu.wpi.first.units.measure.AngularVelocity;
+import org.littletonrobotics.junction.AutoLog;
+
+@AutoLog
+public class XCANMotorControllerInputs {
```
aschokking, line 8: ❓ This is maybe a dumb question, but when I read motor controller inputs, I'm thinking things like desired voltage or target speed. Is there another name for these that could distinguish that it's not really the inputs to the motor controller at a meta level but more like inputs to it's outputs?

PR #491, opened 2025-01-18
Sim Camera streams

aschokking:
# Why are we doing this?

Allow the simulated cameras to stream what they see, making it easier to debug what's going on.

The cameras stream on the normal photonlib ports (starting at 1181)

# How this was tested
image

- [ ] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/subsystems/vision/AprilTagVisionIOPhotonVisionSimulated.java
```diff
@@ -64,6 +64,9 @@ public AprilTagVisionIOPhotonVisionSimulated(@Assisted String name, @Assisted Tr
         // Add sim camera
         var cameraProperties = new SimCameraProperties();
         cameraSim = new PhotonCameraSim(camera, cameraProperties);
+        cameraSim.enableRawStream(true);
+        cameraSim.enableProcessedStream(true);
+        cameraSim.enableDrawWireframe(true);
```
aschokking, line 69: It's possible this will prove too resource intensive on slower machines and then we can come up with more of a dynamic, preferences based config. But in the short-run I think this will help with debugging.

PR #493, opened 2025-01-18
Add PIDProperties to the XCANMotor

JohnGilb:
# Why are we doing this?
- This allows us to rapidly tune PID(F) parameters at runtime for motors.
# Whats changing?
- Porting logic from XCANSparkMAX to th enewer XCANMotor

# How this was tested
- [x] unit tests added
- [x] tested in simulation


stephenjust (approved):

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -31,12 +32,25 @@ XCANMotorController create(
     protected XCANMotorControllerInputsAutoLogged inputs;
     protected String akitName;
 
+    protected boolean usesPropertySystem = true;
```
aschokking, line 35: ⭐ ⭐ could you add comments about what these do and when to change them?

JohnGilb: Added in a followup commit! :)

PR #499, opened 2025-01-18
Improve CAN Motor Debugging

JohnGilb:
# Why are we doing this?
The XCANMotorController logged its position and velocity, but not its power or current. These are valuable when tuning a PID system.
# Whats changing?
- Adds Unit-based AKIT inputs for Voltage and Current
- Fixes a bug in SwerveSteeringSubsystem related to other units

# How this was tested
- [x] tested on robot


src/main/java/xbot/common/advantage/AKitLogger.java
```diff
@@ -16,7 +16,7 @@ public class AKitLogger {
     public enum LogLevel {
         DEBUG, INFO
     }
-    private static LogLevel globalLogLevel = LogLevel.DEBUG;
+    private static LogLevel globalLogLevel = LogLevel.INFO;
```
aschokking, line 19: Do we want this global change to be the default for now?

src/main/java/xbot/common/subsystems/drive/swerve/SwerveSteeringSubsystem.java
```diff
@@ -301,16 +304,29 @@ public void setMotorControllerPidTarget() {
...
+
             // We can rely on either encoder for the starting position, to get the change in angle. Using the CANCoder
             // position to calculate this will help us to avoid any drift on the motor encoder. Then we just set our
             // target based on the motor encoder's current position. Unless the wheels are moving rapidly, the measurements
             // on each encoder are probably taken close enough together in time for our purposes.
             Angle currentPosition = getBestEncoderPosition();
-            double changeInDegrees = MathUtil.inputModulus(target.minus(currentPosition).in(Degrees), -90, 90);
-            Angle targetPosition = this.motorController.getPosition().plus(Degrees.of(changeInDegrees / degreesPerMotorRotation.get()));
+            Angle angleBetweenDesiredAndCurrent = Degrees.of(MathUtil.inputModulus(target.minus(currentPosition).in(Degrees), -90, 90));
+            aKitLog.record("angleBetweenDesiredAndCurrent-Degrees", angleBetweenDesiredAndCurrent.in(Degrees));
```
aschokking, line 318: 🎉

aschokking (approved):

src/main/java/xbot/common/advantage/AKitLogger.java
JohnGilb: Yes, I think we should get into a pattern where it's always "INFO" in the store and only temporarily "DEBUG" in local testing. Ideally it would be a deploy-time variable (but I couldn't figure out how to make that work in IntelliJ)

PR #502, opened 2025-01-21
Feature/cancoder wpi units

stephenjust:
# Why are we doing this?
We're starting to adopt WPI Units elsewhere in the robot code, so do the same for the CANCoder adapter for consistency.

# Whats changing?
XCANCoder and XAbsoluteEncoder will now return outputs in the form of Angle and AngularVelocity

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking (approved):
Great! John was just saying he wanted to pay down this tech debt of not unit-ifying everything

PR #506, opened 2025-01-22
Field Length Increased

kujo27:
# Why are we doing this?
Fixes Pose2d conversions from Blue Alliance to Red Alliance
# Whats changing?
BasePoseSubsystem fieldXMidpointInMeters 

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/subsystems/pose/BasePoseSubsystem.java
+++ b/src/main/java/xbot/common/subsystems/pose/BasePoseSubsystem.java
@@ -46,14 +46,14 @@ public abstract class BasePoseSubsystem extends BaseSubsystem implements DataFra
     protected BooleanProperty rioRotated;
     protected boolean firstUpdate = true;
     protected double lastSetHeadingTime;
-    public static double fieldXMidpointInMeters = 8.2705;
+    public static double fieldXMidpointInMeters = 8.7785;
 
     public BasePoseSubsystem(XGyroFactory gyroFactory, PropertyFactory propManager) {
         log.info("Creating");
         propManager.setPrefix(this);
         imu = gyroFactory.create();
         this.classInstantiationTime = XTimer.getFPGATimestamp();
-        
+
         // Right when the system is initialized, we need to have the old value be
         // the same as the current value, to avoid any sudden changes later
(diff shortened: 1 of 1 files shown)


src/main/java/xbot/common/subsystems/pose/BasePoseSubsystem.java
```diff
@@ -46,14 +46,14 @@ public abstract class BasePoseSubsystem extends BaseSubsystem implements DataFra
     protected BooleanProperty rioRotated;
     protected boolean firstUpdate = true;
     protected double lastSetHeadingTime;
-    public static double fieldXMidpointInMeters = 8.2705;
+    public static double fieldXMidpointInMeters = 8.7785;
```
aschokking, line 49: ⭐ while we're here, consider turning this into a Distance value of Meters.of instead
```suggestion
    public static Distance fieldXMidpoint = Meters.of(8.7785);

aschokking: Maybe we should also add a comment about what years field this is for? Maybe have a list of past lengths in case someone was running 2024?

aschokking (approved):
good catch!

src/main/java/xbot/common/subsystems/pose/BasePoseSubsystem.java
JohnGilb: Where did we get this new value from? The field is relatively evergreen, so this one surprises me a little.

kujo27: The Game Manual specified the field length as approximately 26ft. 5in. by 57ft. 6(7/8)in. I took the length and converted it to meters then halved it.

JohnGilb: Can you verify this using AdvantageScope? If you drive the robot to 8.7785, is it exactly in the middle of the field?

kujo27: Yes, I verified it

PR #507, opened 2025-01-24
Simulating robot relative driving 

kujo27:
# Why are we doing this?
To allow for simulating robot relative driving in AdvantageScope
# Whats changing?
BaseSwerveDriveSubsytem has another move()

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/subsystems/drive/BaseSwerveDriveSubsystem.java
```diff
@@ -325,6 +326,22 @@ public void move(XYPair translate, double rotate) {
...
 
+    /**
+     * This method allows for simulating robot relative driving on AdvantageScope
+     * @param translate The translation velocity.
+     * @param rotate The rotation velocity.
+     * @param currentPose The robot's current position.
+     */
+    public void move(XYPair translate, double rotate, Pose2d currentPose) {
+        move(translate, rotate, new XYPair());
+        lastRawCommandedRotation = rotate;
```
JohnGilb, line 337: I see this is basically duplicating the code in `fieldOrientedDrive` - that code was placed in there as a short-term hack last year. What we should do, especally now that we want to simulate robot-relative driving, is move this code into the underlying `move` method that both functions share.

JohnGilb (changes requested):

aschokking:
Sounded like we didn't want to do this anymore right?

JohnGilb:
Closing this as we are deprecating this older approach

PR #508, opened 2025-01-24
Simplify use of the april tag vision subsystem, and make sure all the…

stephenjust:
… poses are calculated in refreshDataFrame

# Why are we doing this?
The model of subsystem taken from the Advantage Kit sample only allowed consuming vision data from one place, and the previous implementation of the AprilTagVisionSubsystem was not properly following the DataFrameRefreshable interface.

# Whats changing?
Make sure that all poses are calculated in RefreshDataFrame to avoid race condition in pose consumption.

# How this was tested
- [x] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/subsystems/vision/AprilTagVisionSubsystemCamera.java
```diff
@@ -0,0 +1,162 @@
...
+import edu.wpi.first.wpilibj.Alert;
+import org.littletonrobotics.junction.Logger;
+import xbot.common.advantage.DataFrameRefreshable;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.PropertyFactory;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class AprilTagVisionSubsystemCamera implements DataFrameRefreshable {
```
aschokking, line 15: ⭐ wondering if this needs to have Subsystem in the name?

src/main/java/xbot/common/subsystems/vision/AprilTagVisionSubsystemCamera.java
```diff
@@ -0,0 +1,162 @@
...
+
+    // Multipliers to apply for MegaTag 2 observations
+    private final DoubleProperty linearStdDevMegatag2Factor;
+    private final DoubleProperty angularStdDevMegatag2Factor;
+
+    // Standard deviation multipliers for each camera
+    // (Adjust to trust some cameras more than others)
+    private final DoubleProperty cameraStdDevFactor;
+
+    private final List<Pose3d> tagPoses = new LinkedList<>();
```
aschokking, line 39: ⭐ something I've been wondering about, this list doesn't tell us the tag id of each tag, should we keep a hashmap maybe or some other structure so we know what tag id it was? Eg we're looking for a particular tag.

aschokking (approved):

src/main/java/xbot/common/subsystems/vision/AprilTagVisionSubsystemCamera.java
JohnGilb: Agreed - if it doesn't extend BaseSubsystem, probably shouldn't have Subsystem in the name

JohnGilb (approved):

src/main/java/xbot/common/subsystems/vision/AprilTagVisionSubsystemCamera.java
stephenjust: You *could* just compare with the AprilTagFieldLayout. I've added some helpers to at least indicate if a tag is visible land which camera can see it.

src/main/java/xbot/common/subsystems/vision/AprilTagVisionSubsystemCamera.java
stephenjust: Renamed

PR #512, opened 2025-01-26
Document how DataFrameRefreshable works

stephenjust:
# Why are we doing this?
It's not obvious how DataFrameRefreshable works. Documented the behavior so that we don't need to re-discover in the future.

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking (approved):
I'm not sure how often people see this, but if they do they'll learn a lot hopefully 🤞🏻

aschokking (approved):
Not sure how often people see this, but if they do they'll learn a lot 👍🏻

PR #527, opened 2025-02-04
Unit-based-properties: DistanceProperty proof of concept

aschokking:
# Why are we doing this?

As we're using Unit-ed measures more and more, an awkward seam in our current code is our heavy use of DoubleProperties that could be anything. Right now it's up to the property creator to do a good job of saying what the units it'll be interpreted of as (and students almost never do this without prompting).

I've done an example of DistanceProperty here and if this looks good I'll do the other ones we're using heavily like AngleProperty and VelocityProperty which I imagine will cover most of the bases.

NOTE: I tried doing a generic UnitProperty<U extends Unit> but the typing of Unit.of/in is too loose and I couldn't get it working. Unit.of returns Measure<?> for example.

# Whats changing?

- New DistanceProperty class, bit notable behaviors here are:
  - The measure is inferred from the default value (since we need that anyway)
  - The **measures name is appended to the property name** so that in network tables it'll be obvious what the unit is. Example: a property called "maxDistance" that's in meters should show up in the tables as "maxDistance-in-Meters"

# How this was tested
- [x] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/logging/Pluralizer.java
+++ b/src/main/java/xbot/common/logging/Pluralizer.java
@@ -1,6 +1,6 @@
 package xbot.common.logging;
 
-// logic copilot suggestion code from from apache.commons.text
+// basic logic written by Copilot 
 // CHECKSTYLE:OFF
 public class Pluralizer {
     public static String pluralize(String word) {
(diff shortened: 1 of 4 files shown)


JohnGilb:
Seems pretty reasonable - LGTM!

JohnGilb (approved):

PR #529, opened 2025-02-07
Add the ability to fully reset the scheduler

stephenjust:
# Why are we doing this?
When robot unit tests run the scheduler, it uses a static instance of the CommandScheduler created by WPI Lib. That means that we need to be able to clear stale subsystems out of the scheduler if we want to instantiate another scheduler or we could get unexpected behavior when running multiple tests.

# Whats changing?
Add the ability to fully clean out the CommandScheduler. BaseRobot will do this on initialize just in case it's being run as part of a test, but on a real robot the scheduler will start empty.

# How this was tested
- [x] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/command/BaseRobot.java
+++ b/src/main/java/xbot/common/command/BaseRobot.java
@@ -174,6 +174,7 @@ public abstract class BaseRobot extends LoggedRobot {
         // Get the property manager and get all properties from the robot disk
         propertyManager = injectorComponent.propertyManager();
         xScheduler = injectorComponent.scheduler();
+        xScheduler.reset();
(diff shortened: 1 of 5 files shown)


src/main/java/xbot/common/command/BaseRobot.java
```diff
@@ -174,6 +174,7 @@ protected void initializeSystems() {
         // Get the property manager and get all properties from the robot disk
         propertyManager = injectorComponent.propertyManager();
         xScheduler = injectorComponent.scheduler();
+        xScheduler.reset();
```
aschokking, line 177: ⭐ consider a comment here about why this is important (and that it's mainly for tests?)

aschokking (approved):
good find, pernicious bug from static state

PR #530, opened 2025-02-08
Update isFinished method to return true for setpoint commands

aschokking:
# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/command/BaseSetpointCommand.java
+++ b/src/main/java/xbot/common/command/BaseSetpointCommand.java
@@ -12,6 +12,8 @@ public abstract class BaseSetpointCommand extends BaseCommand {
 
     @Override
     public boolean isFinished() {
-        return false;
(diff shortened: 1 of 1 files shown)


src/main/java/xbot/common/command/BaseSetpointCommand.java
```diff
@@ -12,6 +12,8 @@ public BaseSetpointCommand(SupportsSetpointLock system, SupportsSetpointLock...
 
     @Override
     public boolean isFinished() {
-        return false;
+        // Setpoint commands end instantly by default since the set operation usually
+        // happens in initialize and then nothing else needs to happen
```
stephenjust, line 16: Will this break any WhileTrue button presses? (i.e. something else will change the setpoint if this finishes?)

stephenjust (approved):

PR #531, opened 2025-02-08
Fix derivative bug in PID

kujo27:
# Why are we doing this?
Fixes PID bug where D shouldn't be applied
# Whats changing?
PID, PIDManagerTest

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking (approved):
Looks good! Bonus points for a unit test validating this (that would have failed before)

JohnGilb (approved):

PR #532, opened 2025-02-08
Expose the Swerve motors and encoders as Optional

stephenjust:
# Why are we doing this?
Optional is slightly cleaner as an interface
 
# Whats changing?
Always treat the motor controllers as optional

# How this was tested
- [x] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/subsystems/drive/swerve/SwerveSteeringSubsystem.java
+++ b/src/main/java/xbot/common/subsystems/drive/swerve/SwerveSteeringSubsystem.java
@@ -16,6 +16,7 @@ import xbot.common.advantage.AKitLogger;
 import xbot.common.command.BaseSetpointSubsystem;
 import xbot.common.controls.actuators.XCANMotorController;
 import xbot.common.controls.actuators.XCANMotorControllerPIDProperties;
+import xbot.common.controls.sensors.XAbsoluteEncoder;
 import xbot.common.controls.sensors.XCANCoder;
 import xbot.common.controls.sensors.XCANCoder.XCANCoderFactory;
 import xbot.common.injection.electrical_contract.XSwerveDriveElectricalContract;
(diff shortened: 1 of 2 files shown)


src/main/java/xbot/common/subsystems/drive/swerve/SwerveSteeringSubsystem.java
```diff
@@ -104,10 +104,10 @@ public SwerveSteeringSubsystem(SwerveInstance swerveInstance, XCANMotorControlle
      * Set up status frame intervals to reduce unnecessary CAN activity.
      */
     private void setupStatusFramesAsNeeded() {
-        if (this.contract.areCanCodersReady() && this.encoder.hasResetOccurred()) {
```
aschokking, line 107: ⭐ ⭐  is this has reset occured part not important to check anymore?

src/main/java/xbot/common/subsystems/drive/swerve/SwerveSteeringSubsystem.java
```diff
@@ -352,14 +336,14 @@ public void setMotorControllerPidTarget() {
...
+            aKitLog.record("MotorControllerPosition-Rotations", getMotorController().get().getPosition().in(Rotations));
 
-            Angle targetPosition = this.motorController.getPosition().plus(
+            Angle targetPosition = getMotorController().get().getPosition().plus(
                     Rotations.of(angleBetweenDesiredAndCurrent.in(Degrees) / degreesPerMotorRotation.get())
             );
 
             aKitLog.record("TargetPosition-Rotations", targetPosition.in(Rotations));
-            this.motorController.setPositionTarget(targetPosition, XCANMotorController.MotorPidMode.Voltage, 0);
+            getMotorController().get().setPositionTarget(targetPosition, XCANMotorController.MotorPidMode.Voltage, 0);
```
aschokking, line 346: ⭐  I do like the advice I saw that `get()` should basically never be called and it's better to call the more explicitly scary `orElseThrow()`

aschokking: and/or alternatively, at the top of this block maybe cache the motorController like `var motor = this.motorController.orElseThrow()` and then use that motor value in all these places

aschokking (approved):
Overall I do like that this is more concise

PR #535, opened 2025-02-09
Feature/motor controller scaled position

stephenjust:
# Why are we doing this?
The robot code is littered with unit conversions and scaling factors, which are often done incorrectly. If we can centralize the unit conversion and scaling logic, it means there are fewer opportunities for bugs.

# Whats changing?
Add support for an angle scaling factor in all motor controller types when getting or setting positions. Unscaled versions are added via "raw" versions of the methods.

By default the behavior of existing code does not change.

# How this was tested
- [x] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/controls/actuators/XCANMotorController.java
+++ b/src/main/java/xbot/common/controls/actuators/XCANMotorController.java
@@ -1,10 +1,15 @@
 package xbot.common.controls.actuators;
 
+import edu.wpi.first.units.AngleUnit;
 import edu.wpi.first.units.AngularAccelerationUnit;
+import edu.wpi.first.units.DistanceUnit;
+import edu.wpi.first.units.Measure;
+import edu.wpi.first.units.PerUnit;
 import edu.wpi.first.units.measure.Angle;
 import edu.wpi.first.units.measure.AngularAcceleration;
 import edu.wpi.first.units.measure.AngularVelocity;
 import edu.wpi.first.units.measure.Current;
+import edu.wpi.first.units.measure.Distance;
 import edu.wpi.first.units.measure.Time;
 import edu.wpi.first.units.measure.Velocity;
 import edu.wpi.first.units.measure.Voltage;
@@ -24,6 +29,8 @@ import xbot.common.properties.PropertyFactory;
 
 import java.util.function.Supplier;
 
+import static edu.wpi.first.units.Units.Meters;
+import static edu.wpi.first.units.Units.Rotations;
 import static edu.wpi.first.units.Units.Volts;
 
 public abstract class XCANMotorController implements DataFrameRefreshable {
@@ -86,6 +93,9 @@ public abstract class XCANMotorController implements DataFrameRefreshable {
     private DoubleProperty kMinOutputProp;
 
     private static final org.apache.logging.log4j.Logger log = LogManager.getLogger(XCANMotorController.class);
+
+    protected Measure<? extends PerUnit<DistanceUnit, AngleUnit>> distancePerAngleScaleFactor;
+    protected Measure<? extends PerUnit<AngleUnit, AngleUnit>> angleScaleFactor;
 
     protected Supplier<Boolean> softwareReverseLimit = () -> false;
     protected Supplier<Boolean> softwareForwardLimit = () -> false;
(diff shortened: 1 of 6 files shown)


src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -223,33 +235,212 @@ public void periodic() {
 
     public abstract void setPowerRange(double minPower, double maxPower);
 
-    public Angle getPosition() {
+    public void setDistancePerAngleScaleFactor(Measure<? extends PerUnit<DistanceUnit, AngleUnit>> distancePerAngle) {
```
JohnGilb, line 238: This pair of methods is definitely going to need some javadoc to explain to students why they might want to call this and what the input arguments need to be. Would also suggest including a specific example, like an elevator an/or arm for each function to hammer that home.

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -293,4 +484,17 @@ protected boolean isValidPowerRequest(double power) {
         }
         return true;
     }
+
+    protected <N extends Unit, D extends Unit> Measure<? extends PerUnit<D, N>> invertRatio(Measure<? extends PerUnit<N, D>> ratio) {
+        return ratio.unit().reciprocal().of(1 / ratio.magnitude());
```
JohnGilb, line 489: We either need divide-by-zero protection here, or sanitization on the setting side to ensure that `magnitude` would never be zero.

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -223,33 +235,212 @@ public void periodic() {
...
+     * Get the position reported by the motor controller.
+     * @apiNote Distance per angle scaling factors configured on the motor controller are applied.
+     * @return The position reported by the motor controller.
+     */
+    public Distance getPositionAsDistance() {
+        if (distancePerAngleScaleFactor == null) {
+            log.warn("Distance per angle not set for motor controller {}", akitName);
+            return Meters.zero();
+        }
+        return getRawPosition().timesConversionFactor(distancePerAngleScaleFactor);
```
aschokking, line 265: ⭐ in addition to scaling, we almost always have a 'valueAtZero' offset to subtraction from the raw position, is that something we should maybe bake into this as well? Or are you thinking that we'd start tracking that offset at zero in scaled Distance rather than raw positions? Or should we be using setPosition more to zero encoders that way?

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -223,33 +235,212 @@ public void periodic() {
...
+     */
+    public Angle getPosition() {
+        if (angleScaleFactor == null) {
+            return getRawPosition();
+        }
+        return getRawPosition().timesConversionFactor(angleScaleFactor);
+    }
+
+    /**
+     * Override the position of the motor controller.
```
aschokking, line 281: ⭐ consider giving an example of why someone might want to do this

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -223,33 +235,212 @@ public void periodic() {
...
+     * @param position The target position to set.
+     * @param mode The PID mode to use when setting the target position.
+     * @param slot The PID slot to use when setting the target position.
+     * @apiNote Angle scaling factors configured on the motor controller are applied.
+     */
+    public void setPositionTarget(Angle position, MotorPidMode mode, int slot) {
+        if (angleScaleFactor == null) {
+            setRawPositionTarget(position, mode, slot);
+        } else {
+            setRawPositionTarget(position.timesConversionFactor(invertRatio(angleScaleFactor)), mode, slot);
```
aschokking, line 330: ⭐ maybe a nit, but I think it might be better in these cases to pull out the position setting math into it's own bit of logic and then have a shared call to `setRawPositionTarget` after the `if`, that way if the call ever changes someone can't mess up updating one call but not the other by mistake

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -223,33 +235,212 @@ public void periodic() {
...
+        setRawPositionTarget(position, MotorPidMode.DutyCycle);
+    }
+
+    /**
+     * Set the target position for the motor controller.
+     * @param position The target position to set.
+     * @param mode The PID mode to use when setting the target position.
+     * @apiNote Angle scaling factors configured on the motor controller are ignored.
+     */
+    public void setRawPositionTarget(Angle position, MotorPidMode mode) {
```
aschokking, line 349: ⭐ ⭐ After reading a lot of this code, I keep getting confused about what's a raw position vs scaled position/distance. Especially with everything being Angle. A low hanging fruit improvement could be renaming variables like this `position` to `rawPosition`. Longer term I was wondering if we could use the type system to help differentiate these different Angle's into separate types so you can't use the wrong one?

src/test/java/xbot/common/controls/actuators/CANMotorControllerTest.java
```diff
@@ -81,4 +85,37 @@ public void softwareLimitTests() {
         motor.periodic();
         assertEquals(0, motor.getPower(), 0.001);
     }
+
+    @Test
+    public void testScaleFactors() {
```
aschokking, line 90: 👏🏻

aschokking (approved):

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
stephenjust: SetPosition to move the zero seems like the cleanest way to avoid extra math.

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
stephenjust: Added!

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
stephenjust: Refactored

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
stephenjust: I think I fixed this in all the cases to change the parameter name.

PR #538, opened 2025-02-12
Add functiosn to XCANMotorController to set voltage limits

stephenjust:
# Why are we doing this?
Voltage PID modes on TalonFX don't honor duty-cycle power limits. We need to expose their voltage limits.

# Whats changing?
Expose voltage limits and add additional pid properties.

# How this was tested
- [x] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/controls/actuators/XCANMotorController.java
+++ b/src/main/java/xbot/common/controls/actuators/XCANMotorController.java
@@ -77,13 +77,15 @@ public abstract class XCANMotorController implements DataFrameRefreshable {
     protected boolean usesPropertySystem = true;
     protected boolean firstPeriodicCall = true;
 
-    private DoubleProperty kPProp;
-    private DoubleProperty kIProp;
-    private DoubleProperty kDProp;
-    private DoubleProperty kVelocityFFProp;
-    private DoubleProperty kGravityFFProp;
-    private DoubleProperty kMaxOutputProp;
-    private DoubleProperty kMinOutputProp;
+    protected DoubleProperty kPProp;
+    protected DoubleProperty kIProp;
+    protected DoubleProperty kDProp;
+    protected DoubleProperty kVelocityFFProp;
+    protected DoubleProperty kGravityFFProp;
+    protected DoubleProperty kMaxOutputProp;
+    protected DoubleProperty kMinOutputProp;
+    protected DoubleProperty kMinVoltageProp;
+    protected DoubleProperty kMaxVoltageProp;
 
     private static final org.apache.logging.log4j.Logger log = LogManager.getLogger(XCANMotorController.class);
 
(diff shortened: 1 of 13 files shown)


src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -77,13 +77,15 @@ default XCANMotorController create(
...
-    private DoubleProperty kVelocityFFProp;
-    private DoubleProperty kGravityFFProp;
-    private DoubleProperty kMaxOutputProp;
-    private DoubleProperty kMinOutputProp;
+    protected DoubleProperty kPProp;
+    protected DoubleProperty kIProp;
+    protected DoubleProperty kDProp;
+    protected DoubleProperty kVelocityFFProp;
+    protected DoubleProperty kGravityFFProp;
+    protected DoubleProperty kMaxOutputProp;
```
JohnGilb, line 85: What do you think about having just one set of these (e.g. just voltage, or just power) and setting both at runtime (e.g. power*12 to volts, or volts/12 to power?) Do you think we have enough cases where we need to keep these separate, weighted against the risk that somebody sets one set of properties and forgets about the other?

JohnGilb (approved):

src/main/java/xbot/common/controls/actuators/wpi_adapters/CANSparkMaxWpiAdapter.java
```diff
@@ -247,6 +247,11 @@ public Voltage getVoltage() {
                 this.internalSparkMax.getAppliedOutput() * internalSparkMax.getBusVoltage());
     }
 
+    @Override
+    public void setVoltageRange(Voltage minVoltage, Voltage maxVoltage) {
+        log.warn("setVoltageRange: Voltage range is not supported by SparkMax");
```
aschokking, line 252: ⭐ how would code that wants to be device agnostic handle this, should there maybe be a call to check for what capabilities a motor controller supports so they only try to set what it supports?

aschokking (approved):

src/main/java/xbot/common/controls/actuators/wpi_adapters/CANSparkMaxWpiAdapter.java
stephenjust: Won't fix in this PR, but I'll think about it.

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
stephenjust: I'll use one property for the pid settings. Using 13.8 (max charge voltage) as the multiplier to ensure we don't accidentally lose extra voltage on a fresh battery.

setVoltage and setPower remain separate, I expect those only to be used when needed for some special case anyways.

PR #542, opened 2025-02-15
Reduce CANCoder and TalonFX log spew, add Alert signals for health

stephenjust:
# Why are we doing this?
Logging errors on every sensor update is expensive, and it's difficult to get an overview of faults.

# Whats changing?
Ignore errors when reading status signals on CANCoders and TalonFX, and instead emit an alert when unhealthy.

This doesn't eliminate all of the log spew, just most of it.

# How this was tested
- [ ] unit tests added
- [x] tested on robot

### DIFF
--- a/src/main/java/xbot/common/controls/actuators/XCANMotorController.java
+++ b/src/main/java/xbot/common/controls/actuators/XCANMotorController.java
@@ -128,9 +128,11 @@ public abstract class XCANMotorController implements DataFrameRefreshable {
 
         this.propertyFactory.setPrefix(owningSystemPrefix + "/" + info.name());
         police.registerDevice(DevicePolice.DeviceType.CAN, busId, info.deviceId(), info.name());
-        this.akitName = info.name()+"CANMotorController";
+        this.akitName = info.name()+"/CANMotorController";
 
-        this.unhealthyAlert = new Alert("Motor Controller " + info.deviceId() + " on CAN bus " + busId.toString() +  " is unhealthy", Alert.AlertType.kError);
+        this.unhealthyAlert = new Alert("Motor Controller " + info.deviceId() + " on CAN bus " + busId.toString() +  " (" + owningSystemPrefix + ") is " +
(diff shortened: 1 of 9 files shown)


src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -126,6 +130,8 @@ protected XCANMotorController(
         police.registerDevice(DevicePolice.DeviceType.CAN, busId, info.deviceId(), info.name());
         this.akitName = info.name()+"CANMotorController";
 
+        this.unhealthyAlert = new Alert("Motor Controller " + info.deviceId() + " on CAN bus " + busId.toString() +  " is unhealthy", Alert.AlertType.kError);
```
aschokking, line 133: ⭐ ⭐  What do you think about including `owningSystemPrefix` and `pidPropertyPrefix` here so we don't have to go look these up later?

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
```diff
@@ -194,7 +200,11 @@ private void setPIDFromProperties() {
         }
     }
 
+    public abstract DeviceHealth getHealth();
+
     public void periodic() {
+        unhealthyAlert.set(getHealth() == DeviceHealth.Unhealthy);
```
aschokking, line 206: ❓ if it's unhealthy should we be skipping any of these other periodic steps?

aschokking (approved):

src/main/java/xbot/common/controls/actuators/XCANMotorController.java
aschokking, line 133: Also I just noticed Alerts can take a 3rd arg (in first position) for a 'group' that will allow us to group these in the AS so we show some but not others. Maybe these should all go in a `"DeviceHealth"` group?

aschokking (approved):

PR #544, opened 2025-02-16
Set swerve calibrations for X2i, make configurable via contract

stephenjust:
# Why are we doing this?
Calibration for our serve modules is wrong, but each robot will need different calibrations.

# Whats changing?
Expose gear ratios for swerve drive in the contract. Offload the math to the ratio converters in the motor controller classes.

# How this was tested
- [ ] unit tests added
- [x] tested on simulator


JohnGilb (approved):
Approved, but give us some fudge factor input :)

PR #550, opened 2025-02-18
Make configuration writing more resilient on TalonFX

stephenjust:
# Why are we doing this?
In the Phoenix 6 API, when you update a configuration, any unset value will be cleared unless you first do a blocking call to refresh() to read it from the device. This could be easily prone to bugs and accidentally lost configurations.

# Whats changing?
Cache configuration of TalonFX on startup and always work on the cached object, applying sections of it when necessary.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking (approved):
Makes sense. Definitely needs testing on a real robot

PR #551, opened 2025-02-19
2025-02-17 SCL Bringup Support

JohnGilb:
# Why are we doing this?
* Series of changes to support bringing up elevator
# Whats changing?
* Reverts optimizations to the drive system
* Sets LaserCAN to useful configurations by default (make configurable later)
* Change defaults for SwerveDriveSubsystem PID
* 

# How this was tested
- [ ] unit tests added
- [x] tested on robot


stephenjust:
You shouldn't need most of the reverts here. The actual problematic change for the drive has already been reverted.

stephenjust (changes requested):
The problematic changes to SwerveSteeringSubsystem were already reverted. The other changes have been validated to work as far as I know.

PR #559, opened 2025-02-21
Feature/swerve commits to encoders

JohnGilb:
# Why are we doing this?
Every once in a while, when the robot boots, it gets into a state where one or more swerve modules starts rotating rapidly. A code reboot has fixed this each time. I don't have definitive evidence, but from reading through the code, it looks like we check the CANCoder health on boot, and if it's not healthy, we "lock in" to using the onboard motor controller as as PID. 

In practice, since we haven't tuned that PID, we get crazy rotation. Based on our experiences with motor drift, I'm not convinced this is even a practical fallback; I think at this point we have to commit to effectively all or nothing. If we can't connect to the encoder, then we mark the module degraded and always set 0 power to steering and drive, hoping the other 3 modules can take up the slack.

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/controls/sensors/XCANCoder.java
+++ b/src/main/java/xbot/common/controls/sensors/XCANCoder.java
@@ -13,7 +13,7 @@ public abstract class XCANCoder extends XAbsoluteEncoder {
 
     XCANCoderInputsAutoLogged inputs;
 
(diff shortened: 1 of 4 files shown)


aschokking:
We should be able to see that's what's happened in the logs right?

PR #560, opened 2025-02-21
Pigeon2 - as a XGryo

JohnGilb:
# Why are we doing this?
Allows for the pigeon to be a drop-in replacement for the NavX.
# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/controls/sensors/wpi_adapters/Pigeon2Adapter.java
```diff
@@ -0,0 +1,52 @@
...
+        this.pigeon = new Pigeon2(imuInfo.deviceId(), imuInfo.canBusId().id());
+    }
+
+    public void close() {
+        pigeon.close();
+    }
+
+    @Override
+    public boolean isBroken() {
+        return false;
```
aschokking, line 41: ⭐ this doesn't seem super useful, can't we return something about isConnected?

src/main/java/xbot/common/injection/electrical_contract/IMUInfo.java
```diff
@@ -0,0 +1,6 @@
+package xbot.common.injection.electrical_contract;
+
+import xbot.common.controls.sensors.XGyro;
+
+public record IMUInfo(XGyro.InterfaceType interfaceType, CANBusId canBusId, int deviceId) {
```
aschokking, line 5: ⭐  Seems like depending on if it's a CAN device or not the bus id isn't used, could we have 2 different constructors to handle these 2 cases so we don't have to throw a `null` in there and wonder what it's for?

aschokking (approved):

PR #562, opened 2025-02-23
Simple Interpolator

Kobeeeef:
# Why are we doing this?
Allows for Curves without the robot skipping the curve (ChasePoint waits at the next point until a distance)
# Whats changing?
Copying 3 files modified one file
I did this so it wont bother anyone elses code :)
this does not clash with anything else, it is its own classes
# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/subsystems/drive/SwerveSimpleBezierCommand.java
+++ b/src/main/java/xbot/common/subsystems/drive/SwerveSimpleBezierCommand.java
@@ -1,0 +1,90 @@
+package xbot.common.subsystems.drive;
+
+import edu.wpi.first.math.geometry.Twist2d;
+import xbot.common.command.BaseCommand;
+import xbot.common.logging.RobotAssertionManager;
+import xbot.common.math.XYPair;
+import xbot.common.properties.PropertyFactory;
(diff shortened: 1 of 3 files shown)


JohnGilb (approved):
I'm a little concerned that it looks like this is a bunch of copy-pasted code; meaning if we find and fix a bug in one, we have to remember to do it in the other. Is there any way to collapse these interpolator/logic classes into the existing ones as extra "modes" or something like that rather than have duplicate code?

Marking as approved given time constraints, but do evaluate what you have and see if there's a way to go forward here that's easier to maintain.

Kobeeeef:
> I'm a little concerned that it looks like this is a bunch of copy-pasted code; meaning if we find and fix a bug in one, we have to remember to do it in the other. Is there any way to collapse these interpolator/logic classes into the existing ones as extra "modes" or something like that rather than have duplicate code?
> 
> Marking as approved given time constraints, but do evaluate what you have and see if there's a way to go forward here that's easier to maintain.


Well, mainly due to time constraint I did not want to bother any other of your code with mines incase something I change introduces a new bug so I just copied everything over but I can try to after practice match and make a PR for it. But yes, we should use this for now.

PR #566, opened 2025-02-24
AngleProperty + mutable DistanceProperty

aschokking:
# Why are we doing this?

We have a lot of properties that are Angles, so this will help remove a conversion step. 

Also changing these unit based properties to use mutable units internally so hopefully there's slightly less object churn.

# How this was tested
- [x] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/properties/AngleProperty.java
+++ b/src/main/java/xbot/common/properties/AngleProperty.java
@@ -5,22 +5,22 @@ import org.littletonrobotics.junction.LogTable;
 import org.littletonrobotics.junction.Logger;
 import org.littletonrobotics.junction.inputs.LoggableInputs;
 
-import edu.wpi.first.units.DistanceUnit;
-import edu.wpi.first.units.measure.Distance;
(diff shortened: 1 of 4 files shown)


src/main/java/xbot/common/properties/AngleProperty.java
```diff
@@ -0,0 +1,104 @@
...
+
+        return defaultUnit.of(nullableTableValue);
+    }
+
+    public void set(Angle value) {
+        activeStore.setDouble(key, value.in(defaultUnit));
+        currentValue.mut_replace(value);
+    }
+
+    public void hasChangedSinceLastCheck(Consumer<Angle> callback) {
```
JohnGilb, line 80: :star: given all the shared code, it seems like `hasChangedSinceLastCheck` and `hasChangedSinceLastCheck(callback)` should have one of these calling the other. But it's also just four lines of code, so not a big deal.

JohnGilb (approved):

PR #567, opened 2025-02-26
Add ability to get any tag's target observation

stephenjust:
# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


JohnGilb (approved):
Very easy to call & use!

PR #579, opened 2025-03-07
Add is running Alerts to BaseCommand and CommandGroups

aschokking:
# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/command/BaseCommand.java
+++ b/src/main/java/xbot/common/command/BaseCommand.java
@@ -2,6 +2,7 @@ package xbot.common.command;
 
 import javax.inject.Inject;
 
+import edu.wpi.first.wpilibj.Alert;
 import edu.wpi.first.wpilibj2.command.Command;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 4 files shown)


src/main/java/xbot/common/command/BaseParallelCommandGroup.java
```diff
@@ -0,0 +1,120 @@
...
+      m_runWhenDisabled &= command.runsWhenDisabled();
+      if (command.getInterruptionBehavior() == InterruptionBehavior.kCancelSelf) {
+        m_interruptBehavior = InterruptionBehavior.kCancelSelf;
+      }
+    }
+  }
+
+  @Override
+  public final void initialize() {
+    super.initialize();
```
aschokking, line 73: The only delta is here calling super and same thing in end(). Jerks had to mark it final ✊🏻

src/main/java/xbot/common/command/BaseCommand.java
```diff
@@ -44,7 +47,16 @@ public String getPrefix() {
...
-    public abstract void initialize();
+    public void initialize() {
+        // the name might not be set at construction, so let's update it here
+        this.runningAlert.setText(this.getName());
+        this.runningAlert.set(true);
+    }
+
+    @Override
+    public void end(boolean isInterrupted) {
+        this.runningAlert.set(false);
```
aschokking, line 58: We notably need to make sure that child classes remember to call super. I don't have a non-complicated way of enforcing that

stephenjust:
I missed out on this one, but does it not make more sense to do this on XScheduler itself? Have a pool of alert objects managed by the scheduler and call getName() on the scheduled commands?

PR #589, opened 2025-03-17
Feature/lasercan filter invalid v2

stephenjust:
# Why are we doing this?
LaserCAN sensor can report unreliable values, and it can report stale values due to its slow update frequency.

# Whats changing?
Report the time of the latest measurement, and return an Optional.empty() measurement when the sensor data is unreliable.

# How this was tested
- [ ] unit tests added
- [x] tested on robot

### DIFF
--- a/src/main/java/xbot/common/controls/sensors/XLaserCAN.java
+++ b/src/main/java/xbot/common/controls/sensors/XLaserCAN.java
@@ -1,6 +1,7 @@
 package xbot.common.controls.sensors;
 
 import edu.wpi.first.units.measure.Distance;
+import edu.wpi.first.units.measure.Time;
 import org.littletonrobotics.junction.Logger;
 import xbot.common.advantage.DataFrameRefreshable;
 import xbot.common.controls.io_inputs.LaserCANInputs;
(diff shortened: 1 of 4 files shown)


src/main/java/xbot/common/controls/sensors/wpi_adapters/LaserCANWpiAdapter.java
```diff
@@ -46,10 +52,18 @@ public LaserCANWpiAdapter(
     @Override
     public void updateInputs(LaserCANInputs inputs) {
         LaserCanInterface.Measurement measurement = laserCan.getMeasurement();
-        if (measurement != null) {
+        if (measurement != null && measurement.status == LaserCan.LASERCAN_STATUS_VALID_MEASUREMENT) {
```
stephenjust, line 55: This change is what tripped up the previous PR. The jumps to 0 were being handled as valid data. Instead, report an Optional.empty() when the data is invalid, to disambiguate between 0 and invalid data.

src/main/java/xbot/common/controls/sensors/XLaserCAN.java
```diff
@@ -26,8 +29,15 @@ public XLaserCAN(DevicePolice police, DeviceInfo info, String owningSystemPrefix
         inputs = new LaserCANInputsAutoLogged();
     }
 
-    public Distance getDistance() {
-        return inputs.distance;
+    public Optional<Distance> getDistance() {
```
aschokking, line 32: ⭐ I'm wondering if we still want to expose a way to find out what the previous measurement was even if it's not valid?

stephenjust: I thought about that, but it seems too easy to shoot yourself in the foot with such an option.

PR #591, opened 2025-03-22
Run maintainers in disabled, with Coast action

stephenjust:
# Why are we doing this?
On TeleopInit/AutonomousInit, we initialize a whole bunch of commands all at once, possibly leading to high robot cycle times.

# Whats changing?
Startup the maintainers in Coast when disabled, to make sure at least those don't need to be constructed.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot
- [x]  tested on simulator


aschokking (approved):
Pretty elegant, worth trying in the pits maybe

PR #594, opened 2025-04-01
Adding whether a given target observation should be considered stale

rokadias:
This might be worth getting in before the next competition as it might be a reason that we relied on old data, but just a hypothesis at the moment.

Heads up that I have an auto-formatter that usually cleans up files.

# Why are we doing this?
Seeing if this helps avoid issues where we think we might be getting new camera target and pose data, but really we just have an old camera frame.

# Whats changing?
Two new fields on TargetObservation to keep track of when the camera data was added.

# How this was tested
- [x] unit tests updated
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/subsystems/vision/AprilTagVisionIOPhotonVision.java
+++ b/src/main/java/xbot/common/subsystems/vision/AprilTagVisionIOPhotonVision.java
@@ -18,6 +18,7 @@ import edu.wpi.first.apriltag.AprilTagFieldLayout;
 import edu.wpi.first.math.geometry.Pose3d;
 import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.math.geometry.Transform3d;
+import edu.wpi.first.wpilibj.Timer;
 
 import java.util.HashSet;
 import java.util.LinkedList;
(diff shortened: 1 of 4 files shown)


src/main/java/xbot/common/subsystems/vision/AprilTagVisionIOPhotonVision.java
```diff
@@ -43,7 +44,11 @@ public abstract static class FactoryImpl implements AprilTagVisionIOFactory {
     private static final int[] EMPTY_TAG_IDS_OBSERVATION = new int[0];
     private static final PoseObservation[] EMPTY_POSE_OBSERVATION = new PoseObservation[0];
     private static final TargetObservation[] EMPTY_TARGET_OBSERVATIONS = new TargetObservation[0];
-    private static final TargetObservation EMPTY_TARGET_OBSERVATION = new TargetObservation(0, new Rotation2d(), new Rotation2d(), new Transform3d(), 1);
+    private static final TargetObservation EMPTY_TARGET_OBSERVATION = new TargetObservation(0, 0, new Rotation2d(),
+            new Rotation2d(), new Transform3d(), 1, true);
+    // Using same heartbeat bounce as photonvision:
+    // https://github.com/PhotonVision/photonvision/blob/3c332db4bfe9083fc0311ae71cff92de588939ad/photon-lib/src/main/java/org/photonvision/PhotonCamera.java#L107
+    private static final double HEARTBEAT_DEBOUNCE_SEC = 0.5;
```
JohnGilb, line 51: maybe make this public, so the next time we need to reference it we don't need a copy-paste (I wish PV folks had made this a public constant)

stephenjust (approved):

JohnGilb (approved):

PR #600, opened 2025-04-03
Vision/tune

Kobeeeef:
# Why are we doing this?

add tuning to some scaleback values

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/trajectory/SwerveSimpleBezierLogic.java
```diff
@@ -27,6 +27,7 @@ public class SwerveSimpleBezierLogic {
 
     Logger log = LogManager.getLogger(this.getClass());
     final AKitLogger aKitLog = new AKitLogger("SimpleInterpolator/");
+    private double rotationPrioritizationScaleback = 0.25;
```
stephenjust, line 30: Ideally these constants would be properties

stephenjust (approved):

src/main/java/xbot/common/trajectory/SwerveSimpleBezierLogic.java
Kobeeeef: I did this cause I wanted different scaleback per different commands

PR #636, opened 2026-02-15
Auto register all devices for DataFrameRefreshable

aschokking:
## Why

Previously, devices had to be manually registered with a `List<DataFrameRefreshable>` in `BaseRobot`. 

This required annoying boilerplate where each new subsystem needed to remember register its devices as they were created AND then register the subsystem as a whole in BaseRobot.

## What's changing

**Self-registration via `DataFrameRegistry`**

A new `@Singleton` class `DataFrameRegistry` is introduced. Every device wrapper base class now accepts it via Dagger injection and calls `dataFrameRegistry.register(this)` in its own constructor:

- `XCANMotorController`, `XServo`
- `XGyro`, `XEncoder`, `XCANCoder`, `XAbsoluteEncoder`, `XDutyCycleEncoder`, `XSparkAbsoluteEncoder`, `XDigitalInput`, `XLaserCAN`

Because Dagger constructs dependencies before dependents, devices are always registered (and therefore refreshed) before the higher-level subsystems that read their data — no manual ordering required (that's the hope anyway).

**`BaseRobot` simplified**

`refreshAllDataFrames()` is now just:
```java
deviceDataFrameRegistry.refreshAll();
```
The old `protected List<DataFrameRefreshable> dataFrameRefreshables` field and all manual `registerDataFrameRefreshable()` call sites are gone.

**Swerve drivetrain refresh chain fixed**

`BaseSwerveDriveSubsystem` now self-registers with `DataFrameRegistry`, making its `refreshDataFrame()` cascade (→ `SwerveModuleSubsystem` → motor velocity/position reads) live for the first time. A regression test is included that fails without this fix.

**Gyro registration centralized**

`XGyro` now registers in its own base constructor rather than each adapter duplicating the call. As a side effect, `InertialMeasurementUnitAdapter` now always registers even if AHRS hardware initialization throws — previously a hardware failure silently skipped registration.

**`DataFrameRegistry` uses `LinkedHashSet`**

Replaces an O(n) `ArrayList` + `contains()` duplicate check with a `LinkedHashSet`: O(1) dedup with preserved insertion order.

## Testing

- 310 unit tests, 0 failures
- New regression test `refreshDataFrameUpdatesCurrentSwerveStatesFromMotorController` specifically verifies the swerve refresh chain end-to-end


.vscode/settings.json
Rongrrz, line 1: ⭐ I don't think this is too necessary for the PR

src/main/java/xbot/common/command/DataFrameRegistry.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * A registry for components that implement {@link DataFrameRefreshable}, which allows them to be refreshed on a regular basis by the {@link BaseRobot}.
+ */
+@Singleton
+public final class DataFrameRegistry {
+    // LinkedHashSet gives O(1) duplicate-registration checks while preserving registration order,
+    // which refreshAll() relies on (devices are registered, and thus refreshed, before the
+    // higher-level subsystems that read their freshly-refreshed data).
+    private final Set<DataFrameRefreshable> refreshables = new LinkedHashSet<>();
```
Rongrrz: ⭐ If we rely on refreshables to be ordered via LinkedHashSet, then we should just set the type of LinkedHashSet rather than the more abstract Set.

Rongrrz (approved):
I think we should definitely have unit testing for this where we make sure that all the motors get refreshed without explicitly registering them as they are created, just in case if we missed any cases.

.vscode/settings.json
aschokking: That's fair, the watcherExclude is probably good to add but the color  stuff is from an extension I like that makes the vscode sidebar transparent so it stops covering up so much of the code screen. I'll revert it either way for now.

src/main/java/xbot/common/command/DataFrameRegistry.java
aschokking: It's a good question,  `SequencedSet` would capture that this is ordered without having to say how it's implemented, but it's not in the Java version we're using here (introduced in Java 21). I'll change it to the concrete implementation as you've suggested.

src/test/java/xbot/common/command/DataFrameRegistryTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import xbot.common.advantage.DataFrameRefreshable;
+
+import static org.junit.Assert.assertEquals;
+
+public class DataFrameRegistryTest {
```
aschokking, line 13: @Rongrrz this adds a unit test to validate the refreshes happen in registration order. Changing the registry to an unordered HashSet causes the test to fail.

PR #662, opened 2026-03-20
Fix PoseObservation serialization for AdvantageKit replay

aschokking:
# Why are we doing this?

- log replay is useful, we should have it!
- this blocks that
- fixing replay will come in a later PR

# Whats changing?
Claude:

PoseObservation contained a Matrix<N3, N1> estimatedStdDevs field. AdvantageKit's RecordStruct serializes records by   
  reflecting on each component type and looking for a static struct field. Matrix<N3, N1> doesn't have one, so AKit      
  silently drops estimatedStdDevs during logging and throws errors during replay deserialization.                        
                                                                                                                         
  Fix                                                                                                                    
   
  Replace Matrix<N3, N1> estimatedStdDevs in the PoseObservation record with three separate double fields (stdDev0,      
  stdDev1, stdDev2), which RecordStruct can serialize natively.                                                        
                                                                                                                         
  The change is kept as narrow as possible:                                                                              
  - IO implementations (AprilTagVisionIOPhotonVision, AprilTagVisionIOPhotonVisionEstimator) unpack their Matrix values
  only at the new PoseObservation(...) call site                                                                         
  - AprilTagVisionCameraHelper converts back to Matrix<N3, N1> via VecBuilder.fill() when passing to                   
  VisionPoseObservation                                                                                                  
  - getEstimationStdDevs and all other internal logic continue using Matrix<N3, N1> unchanged  

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


rokadias (approved):
Looks like I forgot to approve.

PR #665, opened 2026-03-25
QoL: Return refreshable with registering

Rongrrz:
# Why are we doing this?
I know that currently we have efforts of automatically registering motors to be a refreshable but this removes some of the pain.

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking (approved):
What do you want to chain it with?

Rongrrz:
> What do you want to chain it with?

This was when I was experimenting registering in the same block as motor creation. so like... register(factory.create(...))

PR #667, opened 2026-03-28
refactor: increment multiple keypoints if needed

trisceptre:
# Why are we doing this?
Not too relevant for this year, but incrementing by one keypoint at a time can cause breakage when there is an abundance of keypoints over a short period of time. Additionally, the maximum distance from chase points was changed to 3 so that we won't have to deal with any freezing logic in the interpolator for now.

# Whats changing?

# How this was tested
- [x] unit tests added
- [ ] tested on robot


src/main/java/xbot/common/trajectory/SimpleTimeInterpolator.java
```diff
@@ -158,24 +158,38 @@ public InterpolationResult calculateTarget(Translation2d currentLocation) {
 
 
         // If the fraction is above 1, it's time to set a new baseline point and start LERPing on the next point
-        if (lerpFraction >= 1 && index < keyPoints.size()-1) {
-            // What was our target now becomes our baseline.
+        while (index < keyPoints.size() - 1) {
```
aschokking, line 153: ⭐ are there any unit tests that could be added to help validate this logic?

src/main/java/xbot/common/trajectory/SimpleTimeInterpolator.java
```diff
@@ -158,24 +158,38 @@ public InterpolationResult calculateTarget(Translation2d currentLocation) {
...
-            lerpFraction = 0;
-            log.debug("LerpFraction is above one, so advancing to next keypoint");
             index++;
 
-            // And set our new target to the next element of the list
             targetKeyPoint = keyPoints.get(index);
-            log.debug("Baseline is now " + baseline.getTranslation2d()
-                    + " and target is now " + targetKeyPoint.getTranslation2d());
-
+            log.info("Advancing to next keypoint, new baseline is {} and new target is {}", baseline.getTranslation2d(), targetKeyPoint.getTranslation2d());
```
aschokking: ❔ would we ever have so many keypoints that this logline would become spammy?

aschokking (approved):

src/main/java/xbot/common/trajectory/SimpleTimeInterpolator.java
trisceptre: Most likely not, the regular Simple Interpolator contains a similar line and there were no complaints about log spamming there, at least not to my knowledge.

PR #670, opened 2026-04-01
Add a flag to disable the expensive vision sim for low end machines

aschokking:
# Why are we doing this?

Since the flag is read on boot the program will have to be restarted to be picked up.

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


rokadias (approved):
Vision team doesn't really use the sim this year for anything intensive, and the entire sim runtime is pretty heavy at the moment.

So good to be able to deactivate it.

PR #671, opened 2026-04-01
Adding finish modes for SwerveSimpleTrajectoryLogic

Rongrrz:
# Why are we doing this?
Added a way for users to customize position and rotation thresholds before finishing such that they don't have to rely on pid. Also allow for setting a custom loose velocity.

# Whats changing?
Additionally just refactored older logic so it looks better and is more maintainable.

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


rokadias (commented):
Note: I can make this work for this year since there's a `BaseDriveWithSimpleBezierCommand` where I can keep consistent loose points through out the two sets of movements.

Though imagine if we moved that into SCL. Would it make sense for it to hold what the loose values are it be properties/values coming from the logic.

I'd like to propose that it be a responsibility of the logic. To stick with a better information hiding principal.

rokadias (approved):

PR #673, opened 2026-05-02
AKitLog constructor without IPropertySupport

Rongrrz:
# Why are we doing this?

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/advantage/AKitLogger.java
+++ b/src/main/java/xbot/common/advantage/AKitLogger.java
@@ -46,6 +46,10 @@ public class AKitLogger {
         this(parent.getPrefix());
     }
 
+    public AKitLogger(Object object) {
+        this(object.getClass().getName() + "/");
+    }
+
     /**
      * Temporarily changes the log level for this logger while running the provided code.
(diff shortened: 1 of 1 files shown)


src/main/java/xbot/common/advantage/AKitLogger.java
```diff
@@ -46,6 +46,10 @@ public AKitLogger(IPropertySupport parent) {
         this(parent.getPrefix());
     }
 
+    public AKitLogger(Object object) {
```
aschokking, line 49: What was the situation you wanted this for? Should it instead ask for a string name?

aschokking (approved):

src/main/java/xbot/common/advantage/AKitLogger.java
Rongrrz: I thought that it was more convenient to make AKitLoggers in situations where the object doesn't have a prefix, where we can just do new AKitLogger(this). However, thinking about it more, this is probably less necessary as the more explicit, already-existing version of new AKitLogger(this.getClass().getName()) also seems more reasonable, though for that one it's a longer line and you'll have to manually append a "/".

PR #674, opened 2026-05-15
Disable network table writing of akit logs when on a real field

aschokking:
# Why are we doing this?

We have a hojillian values being written, which tends to tax the rio and the network.

I saw this line in 254s code which disables writing akit logs when fms is attached. It seems like a good idea in general.

The downside is that during a match we might want to be checking a value and won't be able to. For some crucial things like the current auto command we could be using some explicit overrides here maybe.

# Whats changing?

# How this was tested
- [ ] unit tests added
- [ ] tested on robot

### DIFF
--- a/src/main/java/xbot/common/command/BaseRobot.java
+++ b/src/main/java/xbot/common/command/BaseRobot.java
@@ -98,7 +98,12 @@ public abstract class BaseRobot extends LoggedRobot {
                 if (logDirectory.exists() && logDirectory.isDirectory() && logDirectory.canWrite()) {
                     Logger.addDataReceiver(new WPILOGWriter("/U/logs")); // Log to a USB stick with label LOGSDRIVE plugged into the inner usb port
                 }
-                Logger.addDataReceiver(new NT4Publisher()); // Publish data to NetworkTables
+
+                if (!DriverStation.isFMSAttached()) {
+                    // Publish data to NetworkTables if we're not on a real field
+                    Logger.addDataReceiver(new NT4Publisher());
+                }
+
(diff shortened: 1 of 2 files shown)


Rongrrz (commented):
Not too exact on NetworkTable nuances but here are my thoughts:
- Since this is outgoing data, in theory this shouldn't mess with anything Elastic/Shuffleboard/us changing properties on the NT side.
- I do agree that we could definitely lose out on current auto commands (a very immediate--easy--and inefficient solution imo is with system.out).
- We should also survey the drive team to see how often they utilize the live data (but I suppose not very often).
- Lastly: how would this affect if on the robot side we decide to change double properties with code using .set()? (It shouldn't affect anything, right?)

aschokking:
> Not too exact on NetworkTable nuances but here are my thoughts:
> 
> * Since this is outgoing data, in theory this shouldn't mess with anything Elastic/Shuffleboard/us changing properties on the NT side.
> * I do agree that we could definitely lose out on current auto commands (a very immediate--easy--and inefficient solution imo is with system.out).

Instead of System.out, we can still log to network tables explicitly with things like SmartDashboard.put()

> * We should also survey the drive team to see how often they utilize the live data (but I suppose not very often).

I agree

> * Lastly: how would this affect if on the robot side we decide to change double properties with code using .set()? (It shouldn't affect anything, right?)

This doesn't affect Properties at all, just akitlog record lines

aschokking:
In the future maybe we do add back in a concept of log level for these, so we can have a handful of important ones that always go across

Rongrrz (approved):

PR #677, opened 2026-07-30
Remove use of mutable measures

stephenjust:
WPILib 2027 removes the concept of mutable measures, so now is a good time to clean these up.


aschokking (approved):
The reason we moved to these was to address memory pressure from creating tons and tons of immutable objects along the way. Do we maybe need to do something like mutable ourselves with internal doubles more? Or maybe the memory pressure won't be a thing on the newer hardware?

stephenjust:
> The reason we moved to these was to address memory pressure from creating tons and tons of immutable objects along the way. Do we maybe need to do something like mutable ourselves with internal doubles more? Or maybe the memory pressure won't be a thing on the newer hardware?

The story is that they don't believe there will be memory pressure on the Rasperry Pi compute module.

PR #681, opened 2026-09-19
Use Github Workflow

NormalDuck:
# Why are we doing this?
To use less azure credits

# Whats changing?
Migrated to github workflows to build and upload artifacts 

# How this was tested
- [ ] unit tests added
- [ ] tested on robot


aschokking:
@NormalDuck it looks like CI is failing

